PR #2, opened 2019-02-10
scons py2/py3 compatibility

piyushrpt:
Mapping raw_input to input in python3 for the case when scons is built with python3


eric-gurrola (approved):
This is good until pyhon4 comes out, but we can change it then...

PR #3, opened 2019-02-11
converting range mis-registration to meters in tops stack processor

hfattahi:
This PR includes a bug fixing to convert the estimated range-misregistration in pixels to meters in the estimateRangeMisreg.py.


piyushrpt (approved):
This makes it consistent with topsApp

PR #17, opened 2019-03-08
Working SICD reader for example RS2 data

piyushrpt:
First version of SICD RGZERO stripmap sensor. Tested with limited RS2 data and NGA's sarpy library. 
Issuing PR to let interested parties test more extensively. 
Relies on sarpy to be available to read SICD format data.


hfattahi (approved):
Looks good to me.

PR #18, opened 2019-03-08
Add covariance file output for denseoffsets

juribeparada:
This PR add the covariance output file for denseoffsets (Stripmap and TOPS processing). This simply writes to a file the already calculated covariance values by Ampcor. The generated covariance file includes 3 bands (x, y and xy covariance).

### DIFF
--- a/components/isceobj/StripmapProc/runDenseOffsets.py
+++ b/components/isceobj/StripmapProc/runDenseOffsets.py
@@ -65,7 +65,7 @@ def estimateOffsetField(master, slave, denseOffsetFileName,
     
     objOffset.offsetImageName = denseOffsetFileName + '.bil'
     objOffset.snrImageName = denseOffsetFileName +'_snr.bil'
-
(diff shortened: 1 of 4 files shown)


piyushrpt:
```
        self.snr *= self.scaling_factor
        self.cov1 *= self.scaling_factor
        self.cov2 *= self.scaling_factor
        self.cov3 *= self.scaling_factor
```

I think the scaling of the SNR is a bug and should not be carried over to covariance. We should probably remove the scaling from all of this. This was put in place for using the module with optical data like Landsat/Sentinel-2 where the pixel size is known and scaling pixel offsets to meters needed a simple scale factor.

juribeparada:
I understand, should I remove the scaling only for self.cov* or also for SNR?

piyushrpt:
I think snr as well .. I think this was a bug in the original code.

juribeparada:
OK, scaling_factor removed from cov* and snr

piyushrpt (approved):
Thanks for the contribution

PR #19, opened 2019-03-11
stripmapStack: support ALOS filename that are not started with ALP

yunjunz:
This PR adds support for ALOS compressed file from PIXEL in ERI/Japan, whose filename is not started with "ALP" and with multiple levels of sub-directories. The original ASF ALOS file still works. Detailed changes are:

stripmapStack/compressFile: 
1. support multiple levels of sub-directories for both zip and tar file type, instead of one level previously.

stripmapStack/prepRawALOS: 
1. add new function get_ALOS_ALP_name() to grab the "ALP*" string for both ASF and PIXEL ALOS files.
2. remove duplicated inputDir, outputDir, and rmfile variables, and use inps.inputDir, inps.outputDir and inps.rmfile directly;
3. sort the found compressed files.
4. indentation adjustments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 *~
 *.swp
+*.DS_Store
 __pycache__
 .sconf_temp
 .sconsign.dblite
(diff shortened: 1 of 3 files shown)


hfattahi (approved):
Looks good to me.

piyushrpt (approved):

PR #21, opened 2019-03-12
gdal2isce_xml: bug fixed for print() when sch is NoneType

yunjunz:
This PR fixes a bug in the print() in applications/gdal2isce_xml.py when the input vrt file does not have "INTERLEAVE" and sch is assigned with None.


hfattahi (approved):

dbekaert (approved):
Looks good.

PR #25, opened 2019-03-22
stripmapStack: expose --filter_strength and --dual2single options

yunjunz:
This PR exposes two options in stripmapStack: 

1. --dual2single/--fbd2fbs option for ALOS-1 data in prepRawALOS.py to better support the regular interferometry without ionospheric correction.
2. -f / --filter_strength option for stackStripMap.py.

Both options are initiated with default value the same as previously hardwired, so all previous setting should not be affected.

### DIFF
--- a/contrib/stack/stripmapStack/Stack.py
+++ b/contrib/stack/stripmapStack/Stack.py
@@ -5,7 +5,7 @@
 import os, imp, sys, glob
 import argparse
 import configparser
-import  datetime
+import datetime
 import numpy as np
 import shelve
 import isce
@@ -575,7 +575,7 @@ class run(object):
              configObj.igram = configObj.outDir+'.int'
              configObj.filtIgram = os.path.dirname(configObj.outDir) + '/filt_' + pair[0] + '_'  + pair[1] + '.int'
              configObj.coherence = os.path.dirname(configObj.outDir) + '/filt_' + pair[0] + '_'  + pair[1] + '.cor'
-             configObj.filtStrength = filtStrength
+             #configObj.filtStrength = filtStrength
              configObj.filterCoherence('[Function-2]')
 
              configObj.igram = configObj.filtIgram
--- a/contrib/stack/stripmapStack/prepRawALOS.py
+++ b/contrib/stack/stripmapStack/prepRawALOS.py
@@ -192,7 +192,7 @@ def main(iargs=None):
         
 
     # now generate the unpacking script for all the date dirs
-    dateDirs = glob.glob(os.path.join(inps.inputDir,'2*'))
+    dateDirs = sorted(glob.glob(os.path.join(inps.inputDir,'2*')))
     if inps.outputDir is not None:
         f = open(run_unPack,'w')
         for dataDir in dateDirs:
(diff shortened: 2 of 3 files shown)


hfattahi:
It's fine to let users to choose fbd2fbs for ALOS1 data if they want to. But this can mess up the ionosphere estimation. There needs to be a mechanism which doesn't allow fbs2fbd when ionosphere estimation is desired. Any solution for that?

yunjunz:
Maybe add a small function in stackStripMap.py to raise exception and ask user to change the dual2signle setting if:
1. `run_unPackALOS` exists AND
2. `fbd2fbs` is in it AND
3. workflow is `ionosphere`

The location can be within the `cmdLineParse()` or in the main funciton indicated below:
https://github.com/isce-framework/isce2/blob/79b5ad3a4938124228cc4dc4026bbc784090ab58/contrib/stack/stripmapStack/stackStripMap.py#L326

What do you think?

piyushrpt:
My understanding from looking at the code is that the center frequencies and the bandwidths for the subbands are setup as command line options and that these are absolute numbers? I don't see these being computed in the scripts. Is my interpretation correct?

I see that splitspectrum correctly interprets the metadata with chirp slope and chirp duration - which is updated correctly by fbd2fbs. The issue here that user might set these command line inputs based on fbs and this may not be applicable to fbd data due to its narrower spectra. This logic needs to be setup somewhere - either in stackStripmap or in splitSpectrum to cause it to fail.

The setup might work as is  out of the box if the sub band inputs are setup right but probably needs testing to make sure that the ionosphere part is not affected. 

Either way I think, we need to disable fbs2fbd.

hfattahi:
> My understanding from looking at the code is that the center frequencies and the bandwidths for the subbands are setup as command line options and that these are absolute numbers? I don't see these being computed in the scripts. Is my interpretation correct?

That's right Piyush. splitSpectrum.py allows user-input center frequencies of the sub-bands and the bandwidth of each sub-band. The sub-band center-frequencies are relative to full-band center-frequency in Hz. However, these are all optional and not required. If not provided, then they get computed based on the chirp-slope and duration and the general rule of 1/3 of spectrum for each sub-band:
https://github.com/isce-framework/isce2/blob/master/contrib/stack/stripmapStack/splitSpectrum.py#108


The optional user input propagates to the stack setup:
https://github.com/isce-framework/isce2/blob/master/contrib/stack/stripmapStack/Stack.py#L210

> The issue here that user might set these command line inputs based on fbs and this may not be applicable to fbd data due to its narrower spectra. This logic needs to be setup somewhere - either in stackStripmap or in splitSpectrum to cause it to fail.

Yes that's exactly the issue. But,  I don't think if we have a way to check as the metadata have changed during fbd2fbs and represent the fbs spectrum while the original data was fbd. 



> The setup might work as is out of the box if the sub band inputs are setup right but probably needs testing to make sure that the ionosphere part is not affected.

Yes, without doing anything, the setup would work for a stack whose its 14MHz acquisitions have been upsampled to 28MHz. However, the user needs to carefully choose the sub-band center frequencies and bandwidth based on 14 MHz.

piyushrpt:
Looks good to me if its ok with @hfattahi

hfattahi (approved):

PR #30, opened 2019-04-06
Bare minimum Circle CI

piyushrpt:
Setting up a bare minimum circle CI configuration based on the Dockerfile recipe contributed by HySDS. 

Only tries to install the software and run some simple commands.  Build takes about 8-12 minutes. Could be extended to publish an image to dockerhub. Tested on my fork but permissions need to be set up with circle CI as member of isce-framework organization to enable this. 

Note: There could be a libuuid issue with mdx for the Dockerfile recipe. Need to be confirmed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,14 +4,15 @@ jobs:
     docker:
         - image: hysds/pge-base:latest
           user: root
-    working_directory: /root
     steps:
-      - checkout
+      - checkout:
+          path: /root/project/src 
 
       - run:
           name: Install development tools
           command: |
             set -ex
+            pwd
             yum update -y
             yum groupinstall -y "development tools"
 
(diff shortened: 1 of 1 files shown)


piyushrpt:
@rtburns-jpl @pymonger Also please review this

pymonger (approved):
Looks great. As this develops further we can start using circleci workflows to create different workflows that reuse the defined jobs to build and deploy on any PR merge to the master branch and another one that runs periodically. Here's an example of our circleci config that does that: https://github.com/hysds/hysds-dockerfiles/blob/develop/.circleci/config.yml. I initially set up periodic daily builds but that quickly used up my free credits for the month so I made it weekly. The defined "deploy" job also pushes the images up to dockerhub but I encapsulated that in the script.

BryanRiel:
I'm sure you've tested this and everything works great, but I was wondering about the `-luuid` linker flag during the Scons configuration step. Is that necessary? Usually I end up having to "hide" libuuid.so while building ISCE.

rtburns-jpl:
Looks great and I really like where this is going, good insurance against regression. Don't know much about CircleCI but seems easy enough to add tests for a variety of system configurations. Would like to know more about the use of libuuid.

piyushrpt:
I literally grabbed the dockerfile recipe and that included the libuuid ... was wondering about that too. @pymonger - any reason this is enforced in Scons config file?

pymonger:
@piyushrpt @BryanRiel: My goal at the time was to build ISCE in the image so that a user could run mdx and forward its X11 windows to the docker host. I found that I had to add that flag to SConfigISCE otherwise I get this error building:

```
scons: Reading SConscript files ...
Building with scons from python3
Checking for C header file Python.h... yes
Checking for C header file fftw3.h... yes
Checking for C header file hdf5.h... yes
Checking for C header file X11/Xlib.h... yes
Checking for C header file Xm/Xm.h... yes
Checking for C header file omp.h... yes
Checking for C library hdf5... yes
Checking for C library fftw3f... yes
Checking for C library Xm... no
Could not find: Xm lib for libXm
Error: Install Xm or libXm-dev
Checking for C library Xt... no
Could not find: Xt lib for libXt
Error: Install Xt or libXt-dev
Checking for F include fftw3 ... yes
GDAL version: 2.3.2

Checking for C++ header file gdal_priv.h... yes
Checking for C library gdal... yes
Not all components of ISCE will be installed and can result in errors.
Press Enter to continue.... Ctrl-C to exit
```

Is there a better way to do that?

piyushrpt:
@pymonger This is an issue thats cropped up before. Looks like there are issues with conda's libuuid. Our workaround can be found here: 
https://github.com/piyushrpt/oldLinuxSetup/blob/master/issues.md

Don't know, which is a better solution. If mdx works with your solution, thats probably ok. But I think the linker warns about libuuid version everytime you run it.

pymonger:
@piyushrpt nice. I've created an issue (#31) and will resolve and test it.

piyushrpt:
I would actually like to go ahead and merge this to see if the circle ci connection works. I expect a couple of updates in the next week from the user forum. Might be a good chance to test it. Since we are using the hysds docker for testing, we would have to wait till that gets resolved. I will leave the issue open but probably merge this?

bryanvriel (approved):
Yeah, the libuuid linking is a minor issue we can keep open.

PR #34, opened 2019-05-07
fixed DeprecationWarning in stripmapWrapper and SentinelWrapper

mirzaees:


mirzaees:
- used `import importlib` instead of `import imp` and modified `__import` method based on new module
- changed `config.readfp` to `config.read_file` in line 25

piyushrpt:
@mirzaees This would be a welcome fix. Only one suggestion, issue a PR from a named branch rather than master at your end.

mirzaees:
@piyushrpt Sure thank you

PR #35, opened 2019-05-07
Initial midnight fix

piyushrpt:
Fix for images crossing midnight UTC as reported on the user forums. 
No changes required to Apps. All changes are handled within the python interfaces of the processing modules. 

Might need changes to GPU versions (need to verify) to be consistent.


hfattahi (approved):
The fix looks reasonable to me.

dbekaert (approved):
yep looks good.

PR #38, opened 2019-05-08
modified stripmapStack code to work with UAVSAR_STACK

EJFielding:
Several small changes to read in the UAVSAR SLC stack data (UAVSAR_STACK sensor) and setup processing. Ignores the baselines for UAVSAR_STACK sensor, since those are effectively zero.

### DIFF
--- a/contrib/stack/stripmapStack/prepareUAVSAR_coregStack.py
+++ b/contrib/stack/stripmapStack/prepareUAVSAR_coregStack.py
@@ -67,11 +67,11 @@ def main(iargs=None):
     outputDir = os.path.abspath(inps.output)
 
     #######################################
-    slc_files = glob.glob(os.path.join(inps.input, '*_s'+segment+'_1x1.slc'))
+    slc_files = glob.glob(os.path.join(inps.input, '*_s'+inps.segment+'_1x1.slc'))
     for file in slc_files:
         imgDate = get_Date(file)
         print (imgDate)
-        annFile = file.replace('_s'+segment+'_1x1.slc','')+'.ann'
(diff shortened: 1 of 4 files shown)


contrib/stack/stripmapStack/prepareUAVSAR_coregStack.py
```diff
@@ -14,15 +15,15 @@ def createParser():
...
+    parser = argparse.ArgumentParser(description='Prepare UAVSAR SLC Stack files.')
     parser.add_argument('-i', '--input', dest='input', type=str, required=True,
-            help='directory which has all dates as directories. Inside each date, zip files are expected.')
+            help='directory which has all dates.')
     parser.add_argument('-d', '--dop_file', dest='dopFile', type=str, required=True,
-            help='Doppler file for the stack.')
+            help='Doppler file for the stack. Needs to be in directory where command is run.')
     parser.add_argument('-o', '--output', dest='output', type=str, required=True,
             help='output directory which will be used for unpacking.')
-    parser.add_argument('-t', '--text_cmd', dest='text_cmd', type=str, default='source ~/.bash_profile;'
```
piyushrpt: It might be useful to change the default to ''. Having this command in front is useful when different jobs are being submitted via a queuing system on a cluster. Could be used to setup the right env / entrypoint.

EJFielding: Thanks. I think I overwrote that "--text_cmd" parameter in my editing. Would be good to add it back. Do I need to do that in my pull request?

EJFielding: Sorry, it seems that I switched to a different branch when I committed the last change.

PR #39, opened 2019-05-09
added back command line option in prepareUAVSAR

EJFielding:


piyushrpt:
Looks good to me. Only thing to check would be if the all the inputs needed for StaMPS etc downstream are being generated.

EJFielding:
I concentrated on the initial prepare step and getting the selectPairs part to work. Since the UAVSAR SLC stack is already coregistered with all the baseline and topographic phase corrected, we can skip to the last "run_8_igram" step. I ran this on two UAVSAR stacks, but I didn't yet try to run GIAnT on the results.

I don't know what StaMPS needs.

dbekaert (changes requested):
Hi Eric,

This looks good. Wrt to consistent, both stripmap and sentinel stack processors for the SLC workflow generate have the same folder and naming structure when running the SLC workflow. This allows takign the output of the stack processor in e.g. stamps without making sensor dependent changes.

- The baseline grids are one of those. I suggest making baseline grids which are vrt with just point to 0. The grid are coarse and the idea is that a coarse grid scales can be mapped to the original SLC width/length. Given that they are all zero it would not require writing out more than a zero vrt.
- wrt to the date convention we typically have been using YYYYMMDD this might be an issue for UAVSAR as there can be more acquisitions on the same date, e.g. some of the Hawaii data and those for disaster monitoring. Can you confirm how this is handled at this time. Probably here we might need to enforce to include a time-tag too?

EJFielding:
I am at the ESA Living Planet Symposium this week, so we can talk about this next week.

EJFielding:
I changed the Stack.py code so it writes out baseline files with zero values instead of skipping them. This at least allows running Heresh's prepStripmap4timeseries.py. I am not sure if that is sufficient to run the step that creates the baseline grids.

I am not sure if I have any UAVSAR stacks with two acquisitions on the same day. There might be a few where they re-flew a line because they were out of the tube, but both were added to the stack. If we change the file names, then that will affect all the stacks processed in the future, right?

hfattahi:
Looks fine to me. One question, does these changes require running the script only in a specific directory?

EJFielding:
@hfattahi Good question about the directories. I will check the path handling.

hfattahi (approved):

piyushrpt (approved):

dbekaert (approved):
Changes look good.

PR #43, opened 2019-06-10
OMP threads cap fix

piyushrpt:
Fix for OMP threads cap reported on the user forums. 
The reason behind the cap needs to be explored. 
For now, the cap is in place but the number of threads get reset to the original number reported by the system so that other calls in the same process are not affected.


piyushrpt:
Also added a quick fix for GDAL 3 to Sconstruct as reported on the user forums

hfattahi (approved):

bryanvriel (approved):

PR #50, opened 2019-07-19
Fix for importlib has no attribute util

piyushrpt:
Some installations of python 3.6/3.7 (cause to be determined) do not export util in importlib/\_\_init\_\_.py . This might cause installation / tests to fail. Encountered this on some flavors of 3.7 on conda. 

Stack processor wrappers need to be tested. Helper has been tested.


hfattahi (approved):
Hopefully it will work for the wrappers too.

PR #53, opened 2019-07-29
Updates the sensor: UAVSAR_HDF5_SLC

gshiroma:
- fixes issues with reference UTC (when it includes microseconds);
- updates cast from complex32 to complex64 (some versions require complex64 casting through a "sink" - with statement).

### DIFF
--- a/components/isceobj/Sensor/UAVSAR_HDF5_SLC.py
+++ b/components/isceobj/Sensor/UAVSAR_HDF5_SLC.py
@@ -80,6 +80,16 @@ POLARIZATION = Component.Parameter(
     doc='polarization channel of the UAVSAR slc file to be processed'
 )
 
+class DummySink(object):
+    def write(self, data):
+        pass
+
+    def __enter__(self):
+        return self
+
+    def __exit__(*x):
+        pass
+
 from .Sensor import Sensor
 class UAVSAR_HDF5_SLC(Sensor):
     """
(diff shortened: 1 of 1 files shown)


components/isceobj/Sensor/UAVSAR_HDF5_SLC.py
```diff
@@ -247,9 +263,17 @@ def extractImage(self):
...
-            for ii in range(nLines):
-                ds[ii,:].astype(np.complex64).tofile(fout)
+        # if TypeError is raised (e.g. complex32), force casting to complex64
+        try:
+            _ = ds.dtype
+            sink = DummySink()
+        except TypeError:
+            sink = ds.astype(np.complex64)
+
+        with sink:
```
piyushrpt: Wouldn't the case below always work. 
```
with ds.astype(np.complex64) as sink:
```

Can remove the try/ except etc then?

gshiroma: Yes, but if there's no error, e.g. input is complex64, complex128, or even real (although that would be unexpected for SLCs), we don't need to force casting to complex64..

piyushrpt: True. Unfortunately, I dont think we have mechanisms in place in isce2 to handle SLCs other than complex64.

gshiroma: I see. Do you prefer that I force casting to complex64?

piyushrpt: I think for ISCE2, we have to. Since the XML creation that follows essentially says its a CFLOAT and all the SLC reading uses std::complex<float> without a caster check like we do for DEMs.

gshiroma: Ok, will do that. Thanks!

PR #54, opened 2019-07-29
Updates the sensor: UAVSAR_HDF5_SLC

gshiroma:
- fixes issues with reference UTC (when it includes microseconds);
- updates cast from complex32 to complex64 (some versions require complex64 casting through a "sink" - with statement).


hfattahi (approved):
Looks good to me.

PR #64, opened 2019-09-26
Sentinel1 grd to rtc

MIRSL:
These updates mainly concern the rtcApp.py workflow for Sentinel 1 GRD data. It wouldn't run on my system originally, so I made some changes. Besides that, the main purpose was to add radiometric terrain correction (gamma 0) as an output. I think it works pretty well now, but please let me know any issues with these changes.

### DIFF
--- a/components/isceobj/RtcProc/runNormalize.py
+++ b/components/isceobj/RtcProc/runNormalize.py
@@ -9,12 +9,15 @@ import mroipac
 from .runTopo import filenameWithLooks
 from .runLooks import takeLooks
 import os
+import itertools
 import numpy as np
 from isceobj.Util.decorators import use_api
+from applications import imageMath
 
 logger = logging.getLogger('isce.grdsar.looks')
 
-
+class Dummy:
+    pass
 
 def runNormalize(self):
     '''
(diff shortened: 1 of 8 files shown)


components/isceobj/RtcProc/runNormalize.py
```diff
@@ -18,25 +20,29 @@ def runNormalize(self):
...
+        outname = os.path.join(self._grd.outputFolder, filenameWithLooks('gamma_{0}'.format(pol)+'.img', azlooks, rglooks))
+        maskname = os.path.join(self._grd.geometryFolder, self._grd.slMaskFileName)
+
+        cmd = "imageMath.py --e='a*cos(b_0*PI/180.)/cos(b_1*PI/180.) * (c==0)' --a={beta} --b={inc} --c={mask} -o {out} -t FLOAT -s BIL"
+
+        cmdrun = cmd.format(inc = incname,
+                            beta = inname,
+                            out = outname,
+                            mask = maskname)
+        status = os.system(cmdrun)
```
piyushrpt: Would it be possible to replace the os.system() command with direct calls to imageMath.main().
We are trying to not use system calls within the run methods. If not, we can add that as a separate PR. 

For separating the args and files:

```
class Dummy:
    pass

args = Dummy()
args.equation = 'a*cos(b_0*PI/180.)/cos(b_1*PI/180.) * (c==0)'
args.dtype = np.float32
args.scheme = 'BIL'
args.out = outname


files  = Dummy()
files.a = inname
files.b = incname
files.c =  maskname

imageMath.main(files, args)
```

MIRSL: Thanks Piyush! That's no problem, I will look into that soon ... thanks for letting me know.

MIRSL: Hi Piyush, I just updated the codes to avoid the system call as you suggested. I also slightly modified imageMath.py because I encountered some issues otherwise. Thank you!

piyushrpt:
Looks good to me. Let me know if this works and that you have tested it. We can merge.

MIRSL:
Hi Piyush, I've tested this workflow (rtcApp.py) and it works fine! Thank you!

piyushrpt:
Oh ... can you add a simple test to .circleci/config.yml  just like 

```
rtcApp.py --help --steps
```

https://github.com/isce-framework/isce2/blob/master/.circleci/config.yml#L79-L80

piyushrpt (approved):

PR #66, opened 2019-10-01
Prototype for fast staging of data for cloud testing

piyushrpt:
Simple prototype for staging S1 SLC from ASF and earthdata
1. Copies only the annotation files over
2. Setups VRTs within staged zipfiles to point to original earthdata locations 
3. These VRTs can be used to extract individual bursts for scaling

### DIFF
--- a/contrib/stack/topsStack/stageS1_earthdata.py
+++ b/contrib/stack/topsStack/stageS1_earthdata.py
@@ -1,0 +1,441 @@
+#!/usr/bin/env python3
+
+
+def fetchCookies(infile):
+    '''
+    Derived from ASF's bulk downloader python script.
+    '''
+    import os
+    from urllib.request import build_opener, install_opener, Request, urlopen
+    from urllib.request import HTTPHandler, HTTPSHandler, HTTPCookieProcessor
+    from urllib.error import HTTPError, URLError
+    from http.cookiejar import MozillaCookieJar
+    import netrc
+    import base64
+
+
+    def check_cookie_is_logged_in(jar):
+        '''
+        Return true if logged in already.
+        '''
+        for cookie in jar:
+            if cookie.name == 'urs_user_already_logged':
+                return True
+
+        return False
+
+    def check_cookie(cookie_jar):
+        '''
+        Check if cookie in the jar is any good.
+        '''
+        #Make sure cookie is still valid against a known file
+        file_check = 'https://urs.earthdata.nasa.gov/profile'
+
+        # Apply custom Redirect Hanlder
+        ctx = {}
+        opener = build_opener(HTTPCookieProcessor(cookie_jar), HTTPHandler(), HTTPSHandler(**ctx))
+        install_opener(opener)
+
+        # Attempt a HEAD request
+        request = Request(file_check)
+        request.get_method = lambda : 'HEAD'
+        try:
+            print (" > attempting to download {0}".format(file_check))
+            response = urlopen(request, timeout=30)
+            resp_code = response.getcode()
+            # Make sure we're logged in
+            if not check_cookie_is_logged_in(cookie_jar):
+                 cookie_status = False
+
+        except HTTPError:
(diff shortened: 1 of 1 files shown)


piyushrpt:
@scottyhq - FYI. Let us know if there are any optimizations that are possible. 
Used some of the mechanisms from https://github.com/OSGeo/gdal/issues/1000 - you might have played more with these features.

hfattahi (approved):

scottyhq:
This is great to see! I've tried something similar in the past to just pull hh polarization tiffs to the local disk.

Since this streams data via http requests, running on a machine in AWS us-east-1 where the sentinel-1 data is staged would be best. Regarding optimizations I wouldn't be surprised if performance takes a hit compared to simply pulling the entire file with wget or curl. Reading into the zip files with http requests is slow if I recall (especially with the authentication and redirects that are occurring). The benefit of not filling up local storage might very well outweigh the cost though.

There are likely GDAL environment variables that can make a performance difference - see for example https://lists.osgeo.org/pipermail/gdal-dev/2018-December/049507.html. 

Ultimately It's worth putting out there that a great optimization would be to not store zipped SLCs and expose direct /vsis3/ access to the tiffs but that is for another day (... or maybe nisar ;) !

piyushrpt:
@scottyhq - what if we were to process individual bursts on different workers or via dask and each worker only access part of the tiff, without staging a whole tiff. Essentially, make various steps of topsApp run at burst level in a distributed way and only gather them at merge stage? Maybe this is being too optimistic about scalable nature of processing. Currently, the tiffs from ESA are also not chunked up. If these were chunked up, that might also help throughput when pulling out only bits of the tiff. This exercise is to understand implications for NISAR.

scottyhq:
There are definitely two levels of parallelism that could be tested out. 

1) each burst maps to a different worker where an entire topsApp workflow could operate until the merge stage. This could happen on any of the dask cluster variants - [LocalCluster](https://distributed.dask.org/en/latest/local-cluster.html), [PBSCluster](https://docs.dask.org/en/latest/setup/hpc.html), [KubeCluster](https://kubernetes.dask.org/en/latest/) - perhaps with a call to [client.map](https://distributed.dask.org/en/latest/client.html#client). 

Using the vsicurl approach it's possible to pull only the burst data required from a larger tiff as burst1.tif for further processing. Tiling and compression schemes will matter for performance, and it's definitely better to break the full size tiffs out of zip or gz files, just look at the massive log for a simple test of extracting 3 lines:

> CPL_VSIL_CURL_USE_HEAD=YES CPL_DEBUG=ON CPL_CURL_VERBOSE=OFF GDAL_HTTP_COOKIEFILE=asfcookies.txt GDAL_HTTP_COOKIEJAR=asfcookies.txt gdal_translate -srcwin 0 0 13545 3  /vsizip//vsicurl/https://datapool.asf.alaska.edu/SLC/SA/S1A_IW_SLC__1SDH_20190408T113608_20190408T113635_026698_02FF3B_A202.zip/S1A_IW_SLC__1SDH_20190408T113608_20190408T113635_026698_02FF3B_A202.SAFE/measurement/s1a-iw1-slc-hh-20190408t113609-20190408t113635-026698-02ff3b-001.tiff 3lines.tiff

2) if tiffs were stored as cogs, each burst could be read as a dask array and each dask chunk could be operated on in parallel. This may or may not be necessary since a single bust probably fits comfortably in memory. But could be benefitial if operating on a merged product at full resolution https://docs.dask.org/en/latest/array-best-practices.html.

bryanvriel (commented):
This is great @piyushrpt. I tested it out on my laptop and everything worked well. I only have a couple of minor, cosmetic suggestions.

dbekaert (approved):
Tried out the local access via Earthdata and seems good to me.

piyushrpt:
Closing this as ASF's bulk downloader script needs lot more error handling now. This script in its current form no longer works. This would be lot easier if they had a library instead of a monolithing script.

PR #67, opened 2019-10-03
Add files via upload

mgovorcin:


piyushrpt (commented):
1. The unpackframe_ENV looks great. Thanks for the fix. 
2. Can we generalize unpackframe_ERS to either support both ERS_SLC and ERS_Envisat_SLC or have an explicit unpackframe_ERS_ENV? 
3. Do we want to take orbit dir as a command line option to generalize this? I know we had a version that we used for testing earlier and the same error there. Hardcoded paths make it less portable. Might be an opportunity to fix this. You can set defaults to your paths.

mgovorcin:
2.We could generalize it, and I will start working on it. 
3. We could fix the lines with orbit dir by importing the separate user_axu.py file that contains the paths to all auxiliary files (orbit, instrument) per satellite mission. Do you think that contrib folder is the right folder to put the script in, in order to assure the visibility to the script for new users?

piyushrpt:
Sorry for the delayed response. I think there are 2 options:

1. Add a user_aux.py in the stack folder that contains sensor-by-sensor settings - orbits, aux files etc. 

2. Add files to ~/.isce 

I prefer option 1 but can see that it can be an issue if users dont have permissions to edit the stack folder (which is unlikely as they can make a local copy of it and use it).

Piyush

hfattahi:
@piyushrpt @mgovorcin This PR is too old. Shall we close?

PR #68, opened 2019-10-06
stripmapStack. Changing indentations in the baselinerGrid.py

stoormgeo:
Changes to be committed:
	modified:   baselineGrid.py


piyushrpt:
Thank you for the bug fix.

PR #69, opened 2019-10-14
Range rubber sheeting modifications

vbrancat:
1. Separated runRubbersheet.py in runRubbersheetRange.py and runRubbersheetAzimuth.py and added these function as separated steps (rubber_sheet_range, rubber_sheet_azimuth) in stripmapApp.py pipeline
2. When Rubbersheeting in range is turned on

   a. Use range cross-correlation derived offset map to perform the fine coregistration (runResampleSlc.py)
   b. Move flattening at the interferogram formation (runResampleSlc.py and runInterferogram.py). 
   c. Filter the offsets in range and azimuth with new filling algorithm (iterative filling the missing values with a smoothed version of the neighborhood)

### DIFF
--- a/applications/stripmapApp.py
+++ b/applications/stripmapApp.py
@@ -242,14 +242,20 @@ FILTER_STRENGTH = Application.Parameter('filterStrength',
                                       mandatory=False,
                                       doc='')
 
-
-DO_RUBBERSHEETING = Application.Parameter('doRubbersheeting',
-                                      public_name='do rubbersheeting',
+############################################## Modified by V.Brancato 10.07.2019
+DO_RUBBERSHEETINGAZIMUTH = Application.Parameter('doRubbersheetingAzimuth', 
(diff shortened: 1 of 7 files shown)


vbrancat:
Rubbersheet_comparison
Comparison between two ALOS/PALSAR interferograms near Brunt Ice Shelf (Antarctica) processed with rubber sheeting (i.e. speckle-tracking offset map culling) applied only on the azimuth offset map (left interferogram) and with rubber sheeting applied for both the range and azimuth offset maps (right interferogram). The improvement is mainly visible over the floating portion of ice (red rectangle).

hfattahi:
Thanks @vbrancat. This looks great! One minor comment. I think the same changes that you have added to runResampleSlc.py are also needed in runResampleSubbandSlc.py. Would you please add them as well.

hfattahi (approved):
Looks good to me.

bryanvriel (approved):
Cool work and nice results!

PR #76, opened 2019-11-20
cuda: remove arch specific flag; add /lib64 to LIBPATH for libcuda.so…

lijun99:
… for most Linux systems

Some fixed for using CUDA modules:
1. remove -arch=sm_35 which is specific to K40/K80. If no arch is specified, nvcc produces code compatible with sM-30 devices and above. This allows a compatibility with a wider range of GPUs. If a machine specific code is desired, e.g., P100 with sm_60, I recommend to add NVCCFLAGS to SConfigISCE. 
2. libcuda.so comes with nvidia graphic driver and usually is in a system folder such as /lib64.  I added it to cuda.py extension for scons. Otherwise, we could add it CUDALIBPATH in SConfigISCE as well.

### DIFF
--- a/components/zerodop/GPUampcor/cuda/SConscript
+++ b/components/zerodop/GPUampcor/cuda/SConscript
@@ -2,19 +2,19 @@
 
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 # Copyright 2010 California Institute of Technology. ALL RIGHTS RESERVED.
-# 
+#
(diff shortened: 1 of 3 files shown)


lijun99:
Another thing: GPUampcor (Josh's version) uses device cublas routines which are no longer supported by CUDA 10.  Is there a way to detect CUDA version from scons? If so, we can do a conditional include of GPUampcor in isce2/components/zerodop/SConscript. Or, I wonder how many people are using the package; it might be better that we comment out SConscript('GPUampcor/SConscript') for general users.

aivazis:
Josh's version of CUDA ampcor should be removed from the repository. It doesn't work and serves no purpose

piyushrpt:
We dont have a CI system for CUDA stuff. I'm assuming this has been tested.

PR #77, opened 2019-11-20
PyCuAmpcor: updated to the most recent version with gdal input

lijun99:
Updated PyCuAmpcor package. Besides many formatting fixes (indents, comments)..., the major changes are
1) added statistics (by Minyan)
2) now use gdal driver to load images instead of working with the raw slc. Note that the fix to use gdal mmap-based virtualmem was recently merged to gdal github repo. To use this feature, users are encouraged to use the github version of gdal. For old releases, the buffer-based virtualmem is used and might be 40% slower.

### DIFF
--- a/contrib/PyCuAmpcor/SConscript
+++ b/contrib/PyCuAmpcor/SConscript
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import os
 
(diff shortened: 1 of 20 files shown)


piyushrpt:
Again, relying on your word that these changes work. CI system doesnt have a CUDA setup.

PR #78, opened 2019-11-20
PyCuAmpcor: gdalimage.h fixed how to include gdal header

lijun99:
Removed gdal prefix from #include <gdal_priv.h>: only Ubuntu leaves gdal headers to /usr/include/gdal. PyCuAmpcor and other CUDA modules can now compile (with also changes from PR #76), as tested on Ubuntu 18.04 and a RHEL7. PyCuAmpcor is also tested to run properly. I don't know how to test other CUDA modules; will need some help!


lijun99:
I just realized that PyCuAmpcor is included in topsApp/runDenseOffsets. 
1) the new PyCuAmpcor takes ".slc.vrt" files instead of ".slc" directly. Some changes are needed from runDenseOffsets.py.  
2) Do we have an option to choose which GPU to run GPU modules, e.g. _insar.GPUID ?  (This is needed for machines with multiple gpus since we don't want all jobs to compete the default 0 device). 
If it is available, we can use a python function, e.g., PyCuAmpcor.setGPU(_insar.GPUID) to set it.

lijun99:
The new commit changes runDenseOffsets.py and cuDenseOffsets.py, to use master/slave + '.vrt' as inputs to PyCuAmpcor, i.e.,  to add '.vrt' suffix to objOffset.masterImageName.  (my editor seems to remove unnecessary spaces/tabs automatically; there are not many changes).  The topsApp.py with GPU has been tested by Yunjun. 

We still need an option to set GPU device id.

piyushrpt:
Users can also control which devices are visible by just setting CUDA_VISIBLE_DEVICES and then we dont have to worry about setting device ID. This might be ok for isce2. device id handling should be handled better in isce3

piyushrpt (approved):

lijun99:
CUDA_VISIBLE_DEVICES should work. After learning from Cunren how topsApp works, I think it's quite easy to add an Application.Parameter "GPU_ID". I can take a look.

hfattahi (approved):

lijun99:
Can someone merge this PR to master? Thanks!

piyushrpt:
@lijun99  - sorry for the delay. Didnt see that these had been approved.

PR #82, opened 2019-12-14
Hide rubbersheeting scipy/astropy imports

rtburns-jpl:
Recent rubbersheeting changes added scipy and astropy imports which
are not otherwise needed. Due to isce2's eager import behavior this
causes errors when using unrelated functionality. We can fix this
by tucking these imports away inside their function definitions.

To reproduce:
`python3 -c "import isce; from stripmapApp import Insar"`
(without scipy/astropy installed)


piyushrpt (approved):

rtburns-jpl:
Hid the rest of the scipy imports so that stripmapApp.Insar can now be imported without astropy/scipy.

hfattahi (approved):

PR #86, opened 2019-12-17
disable GPUampcor (not used) to allow CUDA >=10.0

lijun99:
GPUampcor requires device cublas libraries which are no longer supported by CUDA 10. Since no one is using this module, it's better we disable it.

### DIFF
--- a/components/zerodop/SConscript
+++ b/components/zerodop/SConscript
@@ -37,6 +37,6 @@ if envzerodop['CYTHON3'] and envzerodop['GPU_ACC_ENABLED']:
     SConscript('GPUtopozero/SConscript')
     SConscript('GPUgeo2rdr/SConscript')
     SConscript('GPUresampslc/SConscript')
-    SConscript('GPUampcor/SConscript')
+#SConscript('GPUampcor/SConscript')
 #SConscript('unw2hgt/SConscript')
 #SConscript('bistaticunw2hgt/SConscript')


rtburns-jpl:
Hmm, are you sure the libraries aren't supported anymore? I think they just changed the package names https://devtalk.nvidia.com/default/topic/1047981/cuda-setup-and-installation/cublas-for-10-1-is-missing/post/5321570/

lijun99:
That's for "host" cublas libraries. Josh used "device" cublas libraries, which call cublas APIs within GPU device functions, and are dropped by NVIDIA since 10.0. See https://docs.nvidia.com/cuda/archive/10.0/cuda-toolkit-release-notes/index.html#deprecated-features under "CUDA Libraries".

rtburns-jpl:
Interesting, thanks for the link. Weird of them to break functionality - do you know the rationale for this change?

lijun99:
No ideas. I guess they take priorities in developing cublas libraries for tensor cores. An alternative is to use batch mode, as I did for my version of cuda ampcor.

piyushrpt:
@lijun99 and @rtburns-jpl - should this be closed now that the other ampcor PR was merged?

lijun99:
This PR is not included in my other PRs. It simply comments out the compilation of GPUampcor
#SConscript('GPUampcor/SConscript')
so that users won't have troubles when compiling ISCE with CUDA10. Since no one is using GPUampcor, I prefer to merge it.

PR #87, opened 2019-12-19
stripmapStack: generate multilooked geometry files

yunjunz:
This PR generates the multilooked geometry files in a new folder "geom_master" from the full resolution geometry files in "merged/geom_master". The detailed changes includes:

Stack:
+ add alks and rlks options in config.topo()

topo: if number of looks is larger than 1, generate multilooked geometry files
+ generate `geom_master` folder in the root directory, same level as `Igrams`
+ import mroipac.looks.Looks module to multilook the full resolution geometry files from `./merged/geom_master`
+ copy over the full resolution xml/vrt files from `./merged/geome` to `./geom_master` as `*.full.xml/vrt` files, to support the number of looks extraction for mintpy
+ bug fixed in extractInfo(). This was not a problem before because the default values are 1 and the custom values from argument was not inputed in the config file.

### DIFF
--- a/contrib/stack/stripmapStack/Stack.py
+++ b/contrib/stack/stripmapStack/Stack.py
@@ -65,6 +65,8 @@ class config(object):
         self.f.write('master : ' + self.slcDir +'\n')
         self.f.write('dem : ' + self.dem +'\n')
         self.f.write('output : ' + self.geometryDir +'\n')
+        self.f.write('alks : ' + self.alks +'\n')
+        self.f.write('rlks : ' + self.rlks +'\n')
         if self.nativeDoppler:
             self.f.write('native : True\n')
         if self.useGPU:
--- a/contrib/stack/stripmapStack/topo.py
+++ b/contrib/stack/stripmapStack/topo.py
@@ -400,6 +400,45 @@ def runMultilook(in_dir, out_dir, alks, rlks):
     return out_dir
 
 
+def runMultilookGdal(in_dir, out_dir, alks, rlks):
+    print('generate multilooked geometry files with alks={} and rlks={}'.format(alks, rlks))
+    import gdal
+
+    # create 'geom_master' directory
+    if not os.path.isdir(out_dir):
+        os.makedirs(out_dir)
+        print('create directory: {}'.format(out_dir))
+
+    # multilook files one by one
+    for fbase in ['hgt', 'incLocal', 'lat', 'lon', 'los', 'shadowMask', 'waterMask']:
+        fname = '{}.rdr'.format(fbase)
+        in_file = os.path.join(in_dir, fname)
+        out_file = os.path.join(out_dir, fname)
+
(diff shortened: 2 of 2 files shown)


hfattahi (commented):
Thank you @yunjunz .
For geometry files, it makes more sense to me to only decimate to the center of the multi-looking window without any averaging/filtering. It does not make much sense to average coordinates.

piyushrpt (approved):

piyushrpt:
I agree with @hfattahi . This becomes easy with odd number of looks in either direction. Dont have a solution other than averaging for even number of looks.

hfattahi:
Perhaps we can go with multi-looking for now and will get back to this later.

hfattahi (approved):

yunjunz:
Hi @hfattahi, I agree that for geometry using the center of the multilook window is more meaningful than averaging. I could check it again with `gdal_translate` as you suggested.

yunjunz:
Hi @hfattahi  and @piyushrpt, I tried 3 methods: a) use `looks.py` from ISCE, b) use 'gdal_translate' with offset set as half of the multilook window int(win_size/2) and c) use 'gdal_translate' with offset set as 0. 

Below are the gdal command I used for Sentinel-1 desc T42 F472 in central San Andreas with alks=5 and rlks=15, with comparison of `hgt.rdr` file between (a) and (b) and between (a) and (c). There are noticeable differences in the mountains. Which one should we use?

```
gdal_translate -of ENVI -srcwin 7 2 45553 11365 -outsize 3037 2273 -a_nodata 0 ../hgt.rdr.full hgt.rdr
```

gdal_translate_offHalfWin

```
gdal_translate -of ENVI -srcwin 0 0 45560 11367 -outsize 3037 2273 -a_nodata 0 ../hgt.rdr.full hgt.rdr
```

gdal_translate_off0

piyushrpt:
I would use a synthetic file where the value in the image is same as line number / pixel number to test. The offset by half window and pixel spacing to looks should reproduce what we expect by mutilooking. 
Will probably play with -tr flag rather than outsize.

yunjunz:
Hi @piyushrpt, I used two synthetic files (for x and y) to test the difference between `looks.py` and `gdal_translate` as you suggested, with the exact same result, as shown below.

x_looks_vs_gdal
y_looks_vs_gdal

The corresponding gdal command is as below and now used in the committed script. 

+ The offset in `-srcwin <xoff> <yoff> <xsize> <ysize>` option has to be zero to get the same results between looks.py and gdal. 
+ `-tr` option can not be used because there is no geotransform info in the input file, thus I still used `-outsize`.

```
gdal_translate -of ENVI -srcwin 0 0 45555 11365 -outsize 3037 2273 -a_nodata 0 ../hgt.rdr.full hgt.rdr
```

Here is the updated comparison between looks.py and gdal for hgt.rdr file.

hgt_looks_vs_gdal

PR #89, opened 2019-12-22
Add CMake build system

rtburns-jpl:
This commit adds experimental CMake build support for ISCE2.

The only changes involved should be adding CMakeLists.txt files.
All current source files, headers, and other functionality, including
the existing SCons build system, should be unaffected.

Some functionality is still a work-in-progress.
These should all be marked as TODO in the CMakeLists.txt files.
`grep TODO **.txt`

Please read CMakeLists.txt and .cmake/*.cmake for more info.


rtburns-jpl:
This should now be synced with ALOS 2 functionality from @CunrenLiang's PRs #107 and #111, as well as scansar removal from #109.

piyushrpt:
Whenever, this is done - think we should update .circleci/config.yml and just repeat build, install and test with the cmake version as well?

rtburns-jpl:
Yeah that sounds good. I noticed some leftover test files while I was setting this up so I added testing commands for them - most are failing but some might be suitable for automation.

contrib/alos2proc_f/src/cbind.f90
```diff
@@ -28,7 +28,8 @@ subroutine c_rect(infile,outfile,ndac,nddn,nrac,nrdn,a,b,c,d,e,f,filetype,intsty
         call rect(infile,outfile,ndac,nddn,nrac,nrdn,a,b,c,d,e,f,filetype,intstyle)
     end subroutine
 
-    subroutine c_rect_with_looks(infile,outfile,ndac,nddn,nrac,nrdn,a,b,c,d,e,f,lac,ldn,lac0,ldn0,filetype,intstyle) bind(c, name="c_rect_with_looks")
+    subroutine c_rect_with_looks(infile,outfile,ndac,nddn,nrac,nrdn,a,b,c,d,e,f,lac,ldn,lac0,ldn0,filetype,intstyle) &
+            bind(c, name="c_rect_with_looks")
```
rtburns-jpl, line 32: Had to wrap this line to avoid fortran -Werror=truncation. For some reason -ffixed-line-length-none did not prevent this.

rtburns-jpl:
Circleci is very cool 😎 

@piyushrpt I think this is ready if you'd like to check it out. The cmake job uses the same base image + deps as the scons job so we might be able to make them share initial steps with some more advanced circleci-fu

piyushrpt:
There have been many changes since this PR started. Do we want to rebase this? 
I also noticed in circleci/config that we are not testing for alos2App and new stuff that have been added. Will add separate PR for that. We can possibly reuse the same for cmake

rtburns-jpl:
I'm fine to rebase/flatten but not sure if that would help. I've tried to keep up the changes as they are merged in but ultimately I think that will only be reliable with more thorough testing.

piyushrpt (approved):

piyushrpt:
@rtburns-jpl merge this in when you feel like this is ready. I will start playing with it this weekend.

PR #93, opened 2020-01-22
Fixes for trackCRs script

rtburns-jpl:
* Strip leading zero from LLH list
* Cast slice bounds to ints
* Use comma delimiter for reading CSVs

### DIFF
--- a/contrib/stack/stripmapStack/trackCRs.py
+++ b/contrib/stack/stripmapStack/trackCRs.py
@@ -73,8 +73,7 @@ def makeOnePlot(filename, pos):
     minx = np.clip(np.min(pos[:,2])-win, 0, npix-1)
     maxx = np.clip(np.max(pos[:,2])+win, 0, npix-1)
 
-
-    box = np.power(np.abs(data[miny:maxy, minx:maxx]), 0.4)
+    box = np.power(np.abs(data[int(miny):int(maxy), int(minx):int(maxx)]), 0.4)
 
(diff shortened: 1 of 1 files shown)


hfattahi (approved):

contrib/stack/stripmapStack/trackCRs.py
```diff
@@ -104,7 +103,7 @@ def getAzRg(frame,llh):
     pol._normRange = frame.instrument.rangePixelSize
     pol.initPoly(azimuthOrder=0, rangeOrder=len(coeffs)-1, coeffs=[coeffs])
 
-    taz, rgm = frame.orbit.geo2rdr(list(llh), side=frame.instrument.platform.pointingDirection, 
+    taz, rgm = frame.orbit.geo2rdr(list(llh)[1:], side=frame.instrument.platform.pointingDirection,
```
piyushrpt: Looks like this essentially changes the format of the text file that the script expects, essentially one more column. Can we add a simple comment indicating the format that the script accepts.

rtburns-jpl: Yeah, these changes were needed for the CSVs providing Rosamond CR data. Is that format atypical?

piyushrpt: Yeah. Its a simple text format - non-standardized. I think it would be good to add what is expected.

PR #95, opened 2020-02-10
Modified Ionospheric Phase correction

vbrancat:
This PR extends the ionospheric phase estimation algorithm present in ISCE2.3 and allows performing the estimation of the dispersive ionospheric phase over grounded ice areas in polar regions. Compared to its previous version, it introduces the following features:

It eliminates the "check_consistency" function. This function aimed at identifying unwrapping phase errors in the low and high band interferograms. In several cases during testing, it was generated artifacts, i.e. spurious phase fringes in the estimated ionospheric phase. We decided to eliminate it.
It introduces a new method for masking outliers during the dispersive component estimation. The already present methods (e.g. coherence thresholding, connected components) poorly perform over grounded ice areas. These areas are generally characterized by medium to high coherence values (0.4-1) and a low fringe rate (i.e. generally easy to unwrap). The newly introduced masking method is based on the procedure described in _H. Liao, F. J. Meyer, B. Scheuchl, J. Mouginot, I. Joughin, E. Rignot, "Ionospheric correction of InSAR data for accurate ice velocity measurements at polar regions", Remote Sensing of Environment, vol. 209, pp. 166-180, May 2018 (H. Liao et al. in the following). The algorithm creates an outlier mask based on the median absolute deviation of the raw dispersive component. The threshold is based on the standard deviation of the dispersive component computed at a mean coherence value selected by the user.
It introduces a new algorithm to fill the holes left after having masked the outliers. The new algorithm (similar to the one reported in runRubbersheetRange.py and runRubbersheetAzimuth.py) fill the holes with smoothed values computed in a neighborhood. Generally, the new filling algorithm has better performances in cases in which the outlier masking leaves big holes. Otherwise, it performs very similarly to the nearest neighbor interpolation (old filling option)
The performance of the newly introduced modifications is tested over several ALOS/PALSAR tracks over grounded ice regions in Greenland and Antarctica also reported in H. Liao et al. .
The picture below shows a comparison of the results obtained with the new ISCE ionospheric phase estimation workflow (first image triplet, in slant range and azimuth geometry) and the results reported in H. Liao et al. (second image triplet, geocoded coordinates) for a ALOS/PALSAR track acquired in the interior of Greenland (i.e. ice velocity below 20 m/year).

Screen Shot 2020-02-10 at 3 42 49 PM


hfattahi:
Thank you @vbrancat. Would you please sync with the master. There seems to be a conflict between master and your repo. In Particular this file:

components/isceobj/StripmapProc/runDispersive.py

It looks like the cv2 import which is now within a "try-except" syntax to let isce install even without cv2.

components/isceobj/StripmapProc/runDispersive.py
```diff
@@ -8,13 +8,10 @@
 from isceobj.Constants import SPEED_OF_LIGHT
 import numpy as np
 import gdal
-<<<<<<< HEAD
 from scipy.ndimage import median_filter
 from astropy.convolution import convolve
-from scipy import ndimage
 import numpy as np
```
rtburns-jpl: This import is duplicated ^

This merge conflict was probably due to https://github.com/isce-framework/isce2/pull/82 where I tried to hide the scipy/astropy imports, is it still possible to import stripmapApp.Insar without these modules? If not can the imports be hidden?

hfattahi (approved):

piyushrpt (approved):

PR #99, opened 2020-03-14
Run master

mirzaees:
I have separated topo calculations and unpacking for master from the rest of slc unpacking in a new run file. This will help better source allocation in parallelization of jobs since topo.py takes long time.
Also parallel processing is implemented in topo.py to process all bursts.

### DIFF
--- a/contrib/stack/topsStack/stackSentinel.py
+++ b/contrib/stack/topsStack/stackSentinel.py
@@ -430,11 +430,16 @@ def slcStack(inps, acquisitionDates, stackMasterDate, slaveDates, safe_dict, upd
     #############################
     i=0
 
+    if not updateStack:
+        i += 1
+        runObj = run()
+        runObj.configure(inps, 'run_' + str(i) + "_unpack_topo_master")
+        runObj.unpackStackMasterSLC(safe_dict)
+        runObj.finalize()
+
     i+=1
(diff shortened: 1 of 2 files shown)


hfattahi:
Thank you @mirzaees for your contribution. I'm curious how much you gain from using multiprocessing within topo.py? Would you please give us an example run time with and without multiprocessing? 
Would you please include, the number of bursts run by topo in your test, the number of cores you have access to, and the number of threads set for OMP_NUM_THREADS.

mirzaees:
Hi @hfattahi, yes  here is a comparison:

number of bursts = 4

With 16 cpu in LSF:
OMP_NUM_THREADS = 4
Sequential = ~24 min
Parallel = ~2 min

OMP_NUM_THREADS = 1
Parallel = ~6 min

With 272 cpu in SLURM (stampede):
OMP_NUM_THREADS = 4
Sequential =  ~35-40 min
Parallel = ~10 min

OMP_NUM_THREADS = 1
Parallel = ~26 min

and I processed 40 bursts in parallel in SLURM and it took ~32 minutes with OMP_NUM_THREADS=4

hfattahi (approved):
Thanks @mirzaees. Looks good to me.

piyushrpt (approved):

PR #102, opened 2020-03-19
dem.py: use abspath filename in xml after correction

yunjunz:
This PR includes the following changes:

+ dem.py: replace filename in xml file with abspath after geoid to ellipsoid correction, so that fixImageXml.py call can be skipped.

+ update README.md files in contrib/stack/*Stack, as fixImageXml.py is not needed anymore

+ fixImageXml: use native argparse mutually exclusive group for --full/--base options to replace checking with own code

### DIFF
--- a/applications/dem.py
+++ b/applications/dem.py
@@ -34,6 +34,8 @@ import sys
 import os
 import argparse
 from contrib.demUtils import createDemStitcher
+
+
 def main():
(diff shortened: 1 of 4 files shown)


dbekaert (approved):
this looks good, makes it easier and was updated for both stack processors.

hfattahi (approved):

PR #103, opened 2020-03-24
stripmapStack/MaskAndFitler: more plotting options

yunjunz:
This PR adds more plotting capabilities in `stripmapStack/MaskAndFilter.py`, including:

+ plot SNR and the evolution of offset after each step to better check the changes by masking, filling and filtering operations.

+ plot both azimuth and range offset in one figure and move the plotting to the end after file writing

+ add `-v, --v-snr, --figsize and --save` options to customize the plot

+ update file path in the XML file so that the script can work even if the files are moved.

It also includes two minors changes in createWaterMask.py and topo.py:

+ createWaterMask: add `--fill` option to be able to change the fill value for the missing waterbody data, because sometimes it's land sometimes it's water.

+ topo: add in_ext/out_ext/fbase_list argument to `runMultilookGdal()` for more flexibility and fix a bug for .full.xml file copying when input/output are in the same directory

### DIFF
--- a/contrib/stack/stripmapStack/MaskAndFilter.py
+++ b/contrib/stack/stripmapStack/MaskAndFilter.py
@@ -4,52 +4,80 @@
 # Copyright 2016
 #
             
-import numpy as np
-import argparse
 import os   
-import isce
-import isceobj
+import argparse
+import numpy as np
+from scipy import ndimage
+import matplotlib.pyplot as plt
 import gdal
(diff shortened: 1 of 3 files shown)


hfattahi (approved):
Looks good to me.

dbekaert (approved):
This looks good to me too.
Like the idea of adding the option to change the fill value, as indeed it can change.

PR #107, opened 2020-03-26
ALOS-2 multi-mode InSAR and ionospheric correction processors

CunrenLiang:
These are the ALOS-2 multi-mode InSAR and ionospheric correction processors. It's important to get it integrated into ISCE rapidly as ARIA system developers and some other users are eager to use them.


piyushrpt:
Hi Cunren,
    Thanks for this detailed PR - great job on including examples for different use cases. 
Most of the changes are well contained within their own modules (eg Alos2proc) and workflows (alos2app). 

Two questions:

1. Do we still need the scansarApp or can we remove it?
2. Could you elaborate on the updates to ampcor and what impact the new parameters have? My understanding is that these parameters don't need to be modified for the current / old behavior but these parameters use special values for ScanSAR processing (possibly downsampling in azimuth due to high oversampling?)

Piyush

CunrenLiang:
Hi Piyush,

I think scansarApp can be removed now. I incorporated some of your codes in scansarApp such as azimuth FM rate computation and others that I think are better than old versions. Thanks for these new routines, I find they are great:-)

Regarding ampcor, if a program calling it sets the following two parameters, the new ampcor will use full-aperture image matching method; otherwise the new ampcor program still use the original matching method.
                #triggering full-aperture mode matching
                ampcor.setWinsizeFilt(8)
                ampcor.setOversamplingFactorFilt(64)
These are two new parameters I added to the fortran and python programs. Since all current ISCE programs don’t set these parameters, so they are not affected and therefore no updates to them are needed.

When matching full-aperture images, the new ampcor program first low-pass filter the image to suppress spikes to avoid false offsets. This is very important for robustness and efficiency. I think no other software packages solved this problem. Incorrect fringes as a result of image spikes occasionally happen. Attached figure shows how false offset affects full-aperture interferogram fringes. I just processed a stack of ~80 ScanSAR interferograms and the new ampcor works perfectly.

Cunren

piyushrpt:
Great. Thanks for the detailed explanation. I think the attachment didnt make it as it was sent by email. Could you upload it here via the github page. I think it will be good to have that for documenting the change.

piyushrpt:
Also, should isceobj/Sensor/ScanSAR be removed as these features are now part of Sensor/MultiMode. We can bring that folder back when we have a true burst-wise ScanSAR sensor that we support.

CunrenLiang:
OK, here is the figure.
[coreg_error_Liang_2018_TGRS.pdf](https://github.com/isce-framework/isce2/files/4385387/coreg_error_Liang_2018_TGRS.pdf)

Yes, I think that folder can also be removed. We do have the burst-wise processor for ALOS-2 ScanSAR. After working with the two workflows for a long time, I think the full-aperture workflow is not necessarily worse than burst-wise processing considering the ionospheric effects.

dbekaert:
Pulling in this version and doing a test for scansar processing.

dbekaert (changes requested):
@CunrenLiang thanks for this contribution. Found it to work well. Speed is also significantly faster compared to the first version, with a lot disk usage. Below some feedback and some issues experienced.


## 1. Example files.
 I believe the following file is now legacy and can be removed? @piyushrpt Can you confirm.
```
ls src/examples/input_files/master_ALOS2.xml
```
New example files have been added for ALOS-2 in a sub-folder. 

```
ls  src/examples/input_files/alos2
```

## 2. Variations in programs.
It will not be as trivial for a non-expert to figure out which variant of the programs and xml input files one should be using. I found the following file to be instructive, but would benefit from a more descriptive file name. Potentially with some more hints or pointers?
```
ls src/examples/input_files/alos2/example_input_files/software_test.txt
```


## 3. Issues:
### A. When  I set:
```
    <property name="number of range offsets for slc matching">5</property>                                                                                        
    <property name="number of azimuth offsets for slc matching">2</property>
```
getting following error:
```
  File "/home/dbekaert/tools/isce/isce-new-ALOS2/install/isce/components/isceobj/Alos2Proc/runSlcOffset.py", line 144, in runSlcOffset
    numberOfOffsetsRange = self.numberRangeOffsets[i][j]
TypeError: 'int' object is not subscriptable
```
Replacing with default, made it run through.

Setting the other multi-look factors options did not causes errors; i.e. for IFG, or iono.


### B. After ISCE install needed to make ALOS-2 related program executable.
```
cd isce/install/applications
chmod +x alos2App.py     alos2burstApp.py
```
This can be fixed by changing the permission of the app to be already executable in the repository.


## 5. Compute time:
Single scansar scene all 5 subswaths with iono correction:
OMP_NUM_THREADS set to  16
total processing time about 2 hours.
Total disk usage at completion ~36GB.


## 6. Questions i got while testing the code
- How do i know if my scansar product has 5 or 7 subswath modes?
- The template xml file in the example provide the default for the multi-look factors as None, but what does None default to in actual value? Can we add these as well as additional information to the xml documentation?  i.e. SLC, ionosphere, IFG

## Output comparison old versus new version.
Figures are much alike.
**new code**
new_code
**old code**
old_code

CunrenLiang:
Thanks for the feedback, David.

> @CunrenLiang thanks for this contribution. Found it to work well. Speed is also significantly faster compared to the first version, with a lot disk usage. Below some feedback and some issues experienced.
> ## 1. Example files.
> 
> I believe the following file is now legacy and can be removed? @piyushrpt Can you confirm.
> 
> ```
> ls src/examples/input_files/master_ALOS2.xml
> ```
> 
> New example files have been added for ALOS-2 in a sub-folder.
> 
> ```
> ls  src/examples/input_files/alos2
> ```
> 
I guess it is for other apps.

> ## 2. Variations in programs.
> 
> It will not be as trivial for a non-expert to figure out which variant of the programs and xml input files one should be using. I found the following file to be instructive, but would benefit from a more descriptive file name. Potentially with some more hints or pointers?
> 
> ```
> ls src/examples/input_files/alos2/example_input_files/software_test.txt
> ```
> 
Users don't have to think about the acquisition mode, settings are almost the same for all modes. Everything is automatically handled by the program. Only when users have an issue or confused by settings, can they refer to these example input files.

> ## 3. Issues:
> ### A. When I set:
> 
> ```
>     <property name="number of range offsets for slc matching">5</property>                                                                                        
>     <property name="number of azimuth offsets for slc matching">2</property>
> ```
> 
> getting following error:
> 
> ```
>   File "/home/dbekaert/tools/isce/isce-new-ALOS2/install/isce/components/isceobj/Alos2Proc/runSlcOffset.py", line 144, in runSlcOffset
>     numberOfOffsetsRange = self.numberRangeOffsets[i][j]
> TypeError: 'int' object is not subscriptable
> ```
> 
> Replacing with default, made it run through.
> 
> Setting the other multi-look factors options did not causes errors; i.e. for IFG, or iono.

These are not number of looks. These are number of offsets in SLC matching. They are 2-d lists. Normally users don't have to set them. The program automatically set number of offsets to use considering both efficiency and robustness.

 
> ### B. After ISCE install needed to make ALOS-2 related program executable.
> 
> ```
> cd isce/install/applications
> chmod +x alos2App.py     alos2burstApp.py
> ```
> 
> This can be fixed by changing the permission of the app to be already executable in the repository.

Hmm, this is a bit strange. I don't have such problem on both mac and linux. Can you help to double-check it?

> ## 5. Compute time:
> 
> Single scansar scene all 5 subswaths with iono correction:
> OMP_NUM_THREADS set to 16
> total processing time about 2 hours.
> Total disk usage at completion ~36GB.

If you have 16 cores, I think do three 5-thread concurrent tasks should be faster if you want to process a stack?

> ## 6. Questions i got while testing the code
> 
>     * How do i know if my scansar product has 5 or 7 subswath modes?
> 
Mode is indicated in the product file names. E.g. IMG-HH-ALOS2046743055-150405-WBDR1.1__D-F1, WBD is the acquisition mode. Users should already know what kind of data they have when ordering the data.  7-subswath mode is not usually used by ALOS-2. Here is the list of acquisition modes:
DDD: Observation Mode
SBS: Spotlight mode
UBS: Ultra-fine mode Single polarization
UBD: Ultra-fine mode Dual polarization
HBS: High-sensitive mode Single polarization
HBD: High-sensitive mode Dual polarization
HBQ: High-sensitive mode Full (Quad.) polarimetry
FBS: Fine mode Single polarization
FBD: Fine mode Dual polarization
FBQ: Fine mode Full (Quad.) polarimetry
WBS: Scan SAR nominal [14MHz] mode Single polarization
WBD: Scan SAR nominal [14MHz] mode Dual polarization
WWS: Scan SAR nominal [28MHz] mode Single polarization
WWD: Scan SAR nominal [28MHz] mode Dual polarization
VBS: Scan SAR wide mode Single polarization
VBD: Scan SAR wide mode Dual polarization

Anyway, users don't have to deals with these, they are all automatically handled by the program.

>     * The template xml file in the example provide the default for the multi-look factors as None, but what does None default to in actual value? Can we add these as well as additional information to the xml documentation?  i.e. SLC, ionosphere, IFG
> 
> 
The default number of looks depends on acquisition mode. They are all indicated in the output filenames. 

I have to think about if we should put all of the default values in the example input file, as these are a lot of values.

> ## Output comparison old versus new version.
> 
> Figures are much alike.
> **new code**
> new_code
> **old code**
> old_code

piyushrpt:
I would leave master_ALOS2.xml as stripmapApp.py can use ALOS2 stripmap data. If any other mode is provided, it raises an error as this only talks to Sensors.ALOS2 file. 

I would also add the description of modes you provided in the response in one of the alos2 tutorials text files. I would not expect science users to know the difference, they would just plugin the data and unfortunately expect things to work. 

Feel free to push to your master branch any changes you have. That will automatically update the PR.

piyushrpt:
I also 3 executables with non-execute permissions:
alos2burstApp.py
alos2App.py
wbd_with_correction.py

dbekaert:
@CunrenLiang, thanks for the detailed feedback.
No further concerns beyond the executable and some files suggestions as @piyushrpt  recommended.

dbekaert:
@CunrenLiang, the documentation changes look great and very helpful.
For the permission commit 139 files changed.
Recommend to only change the permission for the three application files.

CunrenLiang:
Yes, now there are two new changes.

1. tutorial and example input files are updated. I don't think anybody will not be able to setup the input file now. It's pretty simple.
2. executable permission is added to the three applications. 

> @CunrenLiang, the documentation changes look great and very helpful.
> For the permission commit 139 files changed.
> Recommend to only change the permission for the three application files.

The change is not applied to all 139 code files associated with ALOS-2, but only to the three applications. The 139 files you see here are all the files to be added to the upstream ISCE.

dbekaert (approved):
Looks good to me.
Thanks for the changes and the contribution!

dbekaert (approved):
by accident undid my approval. Looks good.

piyushrpt (approved):

piyushrpt:
Thanks Cunren for this contribution

CunrenLiang:
Thank you guys for all the useful comments!

Issue #104, opened 2020-03-25
topsStack bug: occasionally  relativeOrbitNumber to track number conversion wrong


falkamelung:
Occasionally, in topsStack, the relativeOrbitNumber is not properly converted to track number in the /master/IW1.xml file (in about 10-20% of cases). Here an example for  descending track 87 covering Kilauea volcano.   

Here the *.SAFE file for 2018-05-05 showing the correct relativeOrbitNumber of 87:
``` 
grep OrbitNumber S1B_IW_SLC__1SDV_20180505T161525_20180505T161552_010788_013B7D_25D2.SAFE/*
S1B_IW_SLC__1SDV_20180505T161525_20180505T161552_010788_013B7D_25D2.SAFE/manifest.safe:            <safe:relativeOrbitNumber type="start">87</safe:relativeOrbitNumber>
```

However, the master/IW1.xml file shows a track number of 41:
```
/scratch/05861/tg851601/KilaueaSenDT87[1104] head -1422 master/IW1.xml | tail -20
                <property name="sensingstop">
                    <value>2018-05-05 16:15:32.858289</value>
                    <doc>UTC time corresponding to last line of burst SLC</doc>
                </property>
                <property name="startingrange">
                    <value>799103.829224447</value>
                    <doc>Slant range to first pixel in m</doc>
                </property>
                <property name="swathnumber">
                    <value>1</value>
                    <doc>Swath number for bookkeeping</doc>
                </property>
                <property name="terrainheight">
                    <value>0.0</value>
                    <doc>Average terrain height used for focusing</doc>
                </property>
                <property name="tracknumber">
                    <value>41</value>
                    <doc>Track number for bookkeeping</doc>
                </property>
```

Weirdly, some slave granules don't show  the correct  track number, whereas others do (the first one shows 41 and the second correctly 87)
```
//login4/scratch/05861/tg851601/KilaueaSenDT87[1056] head -1422  slaves/20180517/IW1.xml | tail -20
                <property name="sensingstop">
                    <value>2018-05-17 16:15:33.369091</value>
                    <doc>UTC time corresponding to last line of burst SLC</doc>
                </property>
                <property name="startingrange">
                    <value>799103.829224447</value>
                    <doc>Slant range to first pixel in m</doc>
                </property>
                <property name="swathnumber">
                    <value>1</value>
                    <doc>Swath number for bookkeeping</doc>
                </property>
                <property name="terrainheight">
                    <value>0.0</value>
[...]

PR #109, opened 2020-03-29
Cleaning up scansar modules

piyushrpt:
* Remove the obsolete scansarApp.py and related components as this was never meant for general scansar. Replaced by new ALOS2 workflows from #107 

* Fix changes to default values in fixImageXml.py introduced in #102 

* Fix S1B track numbering bug reported in #104


hfattahi (approved):

dbekaert:
Identifying if we have some ARIA dependencies which we might need to resolve internal.

dbekaert (approved):
no conflicts

PR #110, opened 2020-04-05
Fix Burst Loading in Estimate Range and Azimuth Misreg

forrestfwilliams:
Within the for loops that iterate over range(minBurst, maxBurst), the maximum burst number files are not loaded because range() is end exclusive. In cases where only one burst in a swath is used, the scripts fail to produce misregistration estimates. This can be fixed for both scripts by adding `maxBurst = maxBurst +1`, and removing `maxBurst = maxBurst - 1` in estimateRangeMisreg.py


piyushrpt (approved):

hfattahi (approved):

piyushrpt:
Thank you for the bug fixes.

PR #111, opened 2020-04-07
minor updates to ALOS-2 tutorial and input files

CunrenLiang:
minor updates to ALOS-2 tutorial and input files to better explain the parameters
alos2App.xml
alos2_tutorial.txt
alos2burstApp.xml


piyushrpt:
What is the difference between wbd_with_correction.py and wbd.py . Why not update wbd.py with features you want? Why create a completely new parallel script?

CunrenLiang:
wbd_with_correction.py was already put into ISCE in last pull request. For many tiles both on land and water, water body data is missing. wbd.py has this problem. wbd_with_correction.py is not an updated version of wbd.py, it's a completely rewritten program and corrects this problem.

piyushrpt:
Can wbd.py be removed then? Can other scripts (if any) that uses wbd.py be updated to use wbd_with_correction.py?

CunrenLiang:
try wbd.py 20 34 29 42 and wbd_with_correction.py 20 34 29 42 to see the difference of the results.
wbd.py can be completely replaced by wbd_with_correction.py
I am not sure if there are any other apps using wbd.py. As far as I can remember David was also bothered by the missing tiles problem.

piyushrpt:
I understand the improvement. It is better to streamline it and update existing scripts / completely replace it rather than add to documents about not using certain scripts. wbd.py also appears to be part InsarProc which is deprecated. It would be good to highlight as part of future PRs if same functionality is being replicated (even if it is improved). Will let us deprecate the older parts.

CunrenLiang:
I can use the name 'dem.py' for wbd_with_correction.py. I think it might be still good to keep the original script, perhaps name it as 'dem_old.py'.
If everybody is OK with this, I can make this change in this pull request.

dbekaert:
@CunrenLiang  do you use the original dem.py here? https://github.com/isce-framework/isce2/blob/50816f88aed80c9ceb652862c6b1e515fb1c7934/components/isceobj/Alos2Proc/runDownloadDem.py#L39 So would need some more edits.

The new program only provide  wbd information and would need to be updated to provide some more information wrt DEM too

In addition there is a seperate wbd.py program too.

If backward compatibility gets lots might need an increment in the release tag not to be minor. 
For example the original dem.py exposed some more ability to control no-data filling, specifying DEM path etc. @piyushrpt  you think it would be worth exposing this if its meant to be a replacement?

CunrenLiang:
It should be 'wbd.py' instead of 'dem.py' in my last reply. For dem.py I used the original version in ISCE. Regarding wbd.py, I think it's fine to replace it with my new wbd_with_correction.py.

dbekaert:
@CunrenLiang ok, that makes more sense!

Has the same input arguments, so i think this would be fine by me to replace it out given its more robust. I will defer to @piyushrpt what the policy is for deprecated or older versions of a program.

CunrenLiang:
Yes, input, snwe, is tjhe same for both scripts.

piyushrpt:
I suggest, we remove wbd.py . I don't think any other script calls it. ALOS2 tutorials explicitly call wbd_with_correction.py. @yunjunz - do any Mintpy tutorials explicit mention wbd.py

yunjunz:
@piyushrpt Yes, there is one example notebook using wdb.py to download water mask. This can be easily adapted if there is a newer version of it.

https://github.com/insarlab/MintPy-tutorial/blob/master/applications/water_mask.ipynb

CunrenLiang:
OK, two options. 1) remove original wbd.py directly. 2) rename wbd_with_correction.py to wbd.py, and rename or remove original wbd.py.

Let me know what you think.

dbekaert:
There might be other users that are used to see it. I would just replace it as the calls remain the same.

CunrenLiang:
Even if we replace it, I would like to keep the original wbd.py and rename it to something like wbd_old.py

piyushrpt:
@CunrenLiang - I would much prefer if there was one script with a flag to replicate the old behavior than carry two of them. I think that is confusing especially since the old one has a more intuitive name like wbd.py and is likely to the be the first one someone may try to use.

CunrenLiang:
1. original wbd.py is deleted.
2. wbd_with_correction.py is renamed to wbd.py, with an added option for using the original wbd.py
3. ALOS-2 tutorials and example input files in
https://github.com/isce-framework/isce2/tree/master/examples/input_files/alos2
are updated.

piyushrpt:
Looks good. Need to remove wbd_with_correction.py needs to be removed from the SConscript file in applications folder.

CunrenLiang:
Right. Done.

PR #112, opened 2020-04-08
Use exist_ok argument to makedirs when appropriate

rtburns-jpl:
@falkamelung noted a race condition in #106 due to non-atomic directory
checking followed by os.makedirs. This pattern exists in many places
in the codebase so I went ahead and replaced the rest of them.
The new exist_ok usage should be terser and more idiomatic.

### DIFF
--- a/components/isceobj/Alos2Proc/runGeocodeOffset.py
+++ b/components/isceobj/Alos2Proc/runGeocodeOffset.py
@@ -35,7 +35,7 @@ def runGeocodeOffset(self):
     demFile = os.path.abspath(self._insar.demGeo)
 
     denseOffsetDir = 'dense_offset'
-    os.makedirs(denseOffsetDirm, exist_ok=True)
+    os.makedirs(denseOffsetDir, exist_ok=True)
     os.chdir(denseOffsetDir)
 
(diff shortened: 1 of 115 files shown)


components/isceobj/Alos2Proc/runGeocodeOffset.py
```diff
@@ -35,8 +35,7 @@ def runGeocodeOffset(self):
     demFile = os.path.abspath(self._insar.demGeo)
 
     denseOffsetDir = 'dense_offset'
-    if not os.path.exists(denseOffsetDir):
-        os.makedirs(denseOffsetDir)
+    os.makedirs(denseOffsetDirm, exist_ok=True)
```
piyushrpt, line 38: typo in variable name

piyushrpt, line 38:
```suggestion
    os.makedirs(denseOffsetDir, exist_ok=True)
```

piyushrpt (approved):
That must have been a pain to fix. I only caught one typo. Hopefully, others are all fine.

CunrenLiang:
I noticed os.makedirs has such an option after I had completed most of the ALOS-2 codes, so I just continued to check if directory exists instead of using exist_ok.

Issue #96, opened 2020-02-26
Update dem.py to use NASADEM


EJFielding:
I understand that the NASADEM was finally moved from provisional to released status. It is now available at its released URL:
https://e4ftl01.cr.usgs.gov/MEASURES/NASADEM_HGT.001/2000.02.11/

It looks like we need a few changes to the `dem.py` code for the new NASADEM file names and the new URL. Has anybody looked at what is required?

SMJMirzadeh:
Dear Prof. Fielding,

Actually, I have also the same problem in downloading the DEM file. When I checked this website (https://wiki.earthdata.nasa.gov/display/EL/How+To+Access+Data+With+cURL+And+Wget) and found that the configuration for the .netrc file should be the same as this website said.

Best regards,
Sayyed

piyushrpt:
We are happy to take PRs with the required changes.

EJFielding:
Before we start working on a Pull Request, we have to decide whether to simply change the `dem.py` code to use NASADEM or to add an option to use that while leaving the default as SRTM v3.
Any suggestions?

piyushrpt:
The code should already work if a demstitcher.xml (or appropriately named xml) is created with the right base url in the folder where processing is being done. I assumed the suggestion was to change default to NASADEM.

dbekaert:
The changes are trivial to do.

1. Add NASA DEM as an option to dem.py - i.e. add it to the input parser woudl make it the default and You update the URL here:

piyushrpt:
I think this is similar to adding a sensor.

1. Add a new DEMStitcher in location that David pointed to. This stitcher handles the file naming weirdness - prefix, postfix, extra tags etc whatever it takes to build the right tile name.
2. Then this is exposed via the __init__.py in demUtils folder - createDemStitcher method. 
3. Then type argument for dem.py (defaults to version3) can be set to nasadem and interface should work

sssangha:
@EJFielding I have made the above suggested updates under the PR "sss_nasadem"

piyushrpt:
Fixed in #118

EJFielding:
Thanks!

PR #118, opened 2020-04-18
NASADEM download now supported through dem.py

sssangha:
To download NASADEM, run dem.py with 'nasadem' specified as an argument for the '-t' flag. Still defaults to v3 SRTM. 

Changes follow guidelines outlined in issue #96


piyushrpt (approved):
Looks good to me. I tested a couple of tiles with -t nasadem flag and the interface seems to work well. 
Thanks @sssangha

dbekaert (approved):
tiles get pulled in, appears to work.

PR #121, opened 2020-04-25
improvements for ALOS-2 ionospheric correction and interferogram filtering

CunrenLiang:
1. use FFTW for psfiltering
2. enhanced merging of subswaths for ALOS-2 ionospheric correction over poor coherence areas

These only affect ALOS-2 processing, so no need for additional review.

### DIFF
--- a/components/isceobj/Alos2Proc/runIonFilt.py
+++ b/components/isceobj/Alos2Proc/runIonFilt.py
@@ -44,7 +44,8 @@ def runIonFilt(self):
     ###################################
     #SET PARAMETERS HERE
     #THESE SHOULD BE GOOD ENOUGH, NO NEED TO SET IN setup(self)
-    corThresholdAdj = 0.85
+    corThresholdAdj = 0.97
+    corOrderAdj = 20
     ###################################
 
     print('\ncomputing ionosphere')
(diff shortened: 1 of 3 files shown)


hfattahi (approved):

piyushrpt:
Any specific reason to create ps_filt1.c instead of updating and improving ps_filt.c ?

CunrenLiang:
New psfilter can use different window size.

piyushrpt:
Again, why cant this be accommodated as an update to an existing module. Do we need have a whole new copy of the code in a different module?

CunrenLiang:
the code is not modified from the ISCE verison. I can remove the one in mroipac/filter, and place this one in mroipac/filter, but I think this requires some changes in other scripts also.

CunrenLiang:
Just took another look of the filter in mroipac/filter, I found it has changed a lot from the roi_pac version including defines and header files. It would take a lot of time to merge the version in alos2filter and mroipac/filter.

piyushrpt:
I'm a little hesitant to have multiple versions of everything - definite recipe for disaster and for things going out of sync. This is not ideal. For future developments, could you see if the functionality already exists in isce and start from that?

piyushrpt (approved):

PR #124, opened 2020-05-02
gdal2isce_xml: move code from main() to sub-function

yunjunz:
This PR moves the functionality code from main() to a new function gdal2isce_xml() so that it can be called in a python way in other scripts, besides using os.system().


dbekaert (approved):
@yunjunz looks good to me. Allows for a more pythonic call.

PR #125, opened 2020-05-02
stripmapStack/topo: call gdal2isce_xml.py inside runMultilook() if xml file missing

yunjunz:
This PR fixes the bug of sometimes missing xml files during runMultilook() with gdal. Detailed changes are as below:

+ topo: merge runMultilookGdal() into runMultilook() with new argument "method" to control to use isce Looks module (averaging) or gdal_translate (extraction; default, same as before).

+ topo: call gdal2isce_xml.py so that xml file for multilooked data are always generate. This fix the warning message from gdal_translate: "Warning 1: Geotransform matrix has non rotational terms"

+ topo: skip multilooking if the data file has vrt/xml file missing, which sometimes happens to incLocal and shadowMask (don't know the cause yet; I used GPU version).

+ crossmul/resampleSlc: re-organize module import at the top of scripts

### DIFF
--- a/contrib/stack/stripmapStack/crossmul.py
+++ b/contrib/stack/stripmapStack/crossmul.py
@@ -1,12 +1,15 @@
 #!/usr/bin/env python3
-import isce
+
+
 import os
-import logging
-from components.stdproc.stdproc import crossmul
-import isceobj
-from iscesys.ImageUtil.ImageUtil import ImageUtil as IU
-import isceobj
 import argparse
+import logging
+
+import isce
+import isceobj
+from components.stdproc.stdproc import crossmul
+from iscesys.ImageUtil.ImageUtil import ImageUtil as IU
+
 
 def createParser():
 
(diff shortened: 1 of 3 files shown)


piyushrpt:
Would it be possible to eliminate the system calls with python calls for gdal_translate with gdal.Translate and gdal2isce_xml with the updates you made?

yunjunz:
Hi @piyushrpt, the new commit is now using `gdal.Translate()`. 

I did not use the pythonic `gdal2isce_xml()` because this function is imported from `applications`.  Considering the `conda install isce2` users (including myself), this function won't be available until the next condo build. Maybe we should wait until then to replace it? If there is another way to import it now, please let me know.

piyushrpt:
I think something like this would always work, even with conda as it essentially follows the python package directory structure.

```
from isce.applications.gdal2isce_xml import gdal2isce_xml 
```

piyushrpt:
I see what you are saying about the python interface not being available in conda from the release. Lets open a ticket for this in issues and close it on next release.

piyushrpt (approved):

hfattahi (approved):

Issue #120, opened 2020-04-23
rtcApp.py is incomplete


piyushrpt:
@MIRSL - looks like the rtcApp.py PR is incomplete. The geocode step seems to be missing and the the example input file seems to have errors in it. 

Could you follow up and complete the missing steps since you already have it working ?

sgk0:
We haven't added the geocode step as part of the rtcApp.py workflow yet. We've been using a geolocation array approach outside of the workflow, as in this example: https://github.com/isce-framework/isce2-docs/blob/master/Notebooks/GDAL/03_RasterProjection.ipynb. I'll try to add this step in the near future 

I'm not sure about errors in the input file, will look into it.

piyushrpt:
@HBaldwin3 - #120 has been merged. Could you let us know if this suffices?
@sgk0 - thanks for looking into this promptly. 

If everything looks ok, I will retire #119

shobhitagarwal1612:
Thank you so much for the fix! I tried the latest changes and it runs geocoding step as expected.

When I tried it with `spacing` 100, I didn't notice any errors/warnings. 
But doing the same with `spacing` 10 generated the following warnings.

```
****Geocoding file 1 out of 2: gamma_VV.10alks_10rlks.img.vrt****
ERROR 1: Too many points (529 out of 529) failed to transform, unable to compute output bounds.
Warning 1: Unable to compute source region for output window 0,0,3564,2676, skipping.
ERROR 1: Too many points (529 out of 529) failed to transform, unable to compute output bounds.
Warning 1: Unable to compute source region for output window 3564,0,3564,2676, skipping.
ERROR 1: Too many points (529 out of 529) failed to transform, unable to compute output bounds.
Warning 1: Unable to compute source region for output window 17821,18732,3564,2676, skipping.
ERROR 1: Too many points (529 out of 529) failed to transform, unable to compute output bounds.
Warning 1: Unable to compute source region for output window 21385,18732,3564,2676, skipping.
ERROR 1: Too many points (529 out of 529) failed to transform, unable to compute output bounds.
Warning 1: Unable to compute source region for output window 24949,18732,3565,2676, skipping.
****Geocoding file 2 out of 2: gamma_VH.10alks_10rlks.img.vrt****
ERROR 1: Too many points (529 out of 529) failed to transform, unable to compute output bounds.
Warning 1: Unable to compute source region for output window 0,0,3564,2676, skipping.
[...]

PR #119, opened 2020-04-21
Change to README & SConscript Files

hbparache:
README: I added "s" to "polarization" in the example rtcApp.xml
Added rtcApp.py to the applications SConscript file
Added runNormalize.py to the rtcProc SConscript file

PR #128, opened 2020-05-03
pull request in response to rtcApp.py is incomplete #120

sgk0:
Added the geocoding step. It just uses gdal and the geolocation arrays as per the data in the geometry folder. This step comes with a few additional required input parameters.

Quick explanation...
"posting" : just for multilooking, integer multiple of spacing. Putting a large value speeds up processing if highest resolutions are not needed.
"epsg id":  need to put epsg code for utm. I only set up geocoding for utm.
"geocode spacing": this is the input parameter for gdal, and sets the x and y spacing. It may be desireable to have this different from posting value, so I left this option.
"geocode interpolation method": any of the gdal supported ones such as 'near' and 'bilinear'


piyushrpt:
This is related to #119 . Would you like to include those changes to SConscript files in this PR or would you prefer that we merge #119 before we update this.

I think runGeocode.py and its inclusion in SConscript is missing in this PR.

sgk0:
Hi Piyush!

From what I saw it wouldn't be a problem to do #119 first before updating
this. And you're correct, I didn't have all the files there. I'm still a
bit new to github and took me a while how to put my old branch up to date.
In any case, they are uploaded again to my master, and they should show up
now.

Best, Simon

piyushrpt (approved):

PR #130, opened 2020-05-05
stackSentinel: use 2 digit number in run_files to better sort 10 or more steps

yunjunz:
This PR changes the number in run files from 1 to 2 digits, so that for more than 10 steps (pretty common to interferogram workflow), the glob search of "run_*" returns in the order of numbers.

It also includes a default value typo fix in FilterAndCoh.py --cc from #97.


yunjunz:
Fix two bugs related with #113 for topsStack and #129 for stripmapStack.

hfattahi (approved):
Looks great! Thank you @yunjunz

dbekaert (approved):
@yunjunz looks good to me. 

@pymonger @hookhua note that this would impact and requires changes for ARIA team.

PR #132, opened 2020-05-12
updates for alos2App.py and alos2burstApp.py

CunrenLiang:
1. improved ionospheric correction
2. geocoding file list accept wild card and update input xml files accordingly
3. fixed a bug (return to main working dir) in runRdrDemOffset.py when there are not enough good matching points.


dbekaert (approved):
@CunrenLiang Looks good to me. Thanks for adding the geocoding support with wild-card. Will make it easier for production.

piyushrpt (approved):

PR #135, opened 2020-05-20
Fall back to MKL if FFTW is not found 

rtburns-jpl:
See https://github.com/conda-forge/isce2-feedstock/issues/17. Wrapper headers added to extern/ subdir since they don't seem to be included with conda MKL.


rtburns-jpl:
Output snippet from pipelined cmake configuration:
```
-- Could NOT find FFTW (missing: FFTW_LIBRARIES FFTW_INCLUDE_DIRS) 
-- FFTW not found. Falling back to Intel MKL wrapper...
-- Found MKL library: /opt/conda/lib/libmkl_rt.so
```

I don't think incorporating this into the scons files will be as straightforward, so unless a scons wizard can lend a hand, it sounds easier to move our feedstock over to cmake.

piyushrpt:
Are we sure that MKL does not lead to similar license issues since its distributed under an intel specific license? 

There also seems to be specific issues with mkl and certain linux distributions like SELinux.  I think we might need to find out more before committing to the switch.
https://docs.anaconda.com/mkl-optimizations/

I think its good to know that there is a solution, but we don't want to force users towards mkl? I would vote for waiting and watching what is decided on the conda thread.

rtburns-jpl:
For sure, feel free to mark this as a draft until further discussion/testing.

Afaict the issue is just the "infectiousness" of the GPL when dylinking, which usually does not come into play when linking with third-party libs under other licenses. Fwiw MKL happens to be under a fairly permissive "Intel Simplified Software License".
https://software.intel.com/content/www/us/en/develop/articles/end-user-license-agreement.html

piyushrpt:
Are there problems with distributing these specific header files? I see language needs to be included in the binary distributions as well (comments in header file). Do you want to bring this up in the weekly discussion and see what others think?

rtburns-jpl:
The fftw3.h header is a special case so we would only need to comply with those BSD-2-clause conditions, which are much more lenient. You're right that we would still need to include those with the distribution.

rtburns-jpl:
Closing as this seems to be a non-issue? Can always reopen if there is interest

PR #136, opened 2020-05-22
CMake bugfixes and documentation

rtburns-jpl:
Fixes some missing components under cmake, and adds some basic cmake usage info to the readme


piyushrpt (approved):

piyushrpt:
PyCuAmpcor might have issues under CUDA 10.x - #86 
There is no way to figure out cuda version on scons side atleast without writing any custom code. 
Maybe we should add this to "TODO" list for cmake for a later date

rtburns-jpl:
Well at least in cmake that's quite easy to handle, see https://github.com/isce-framework/isce2/pull/136/files#diff-265022fce82837adda18497e0e59042cR9. Is this needed for PyCuAmpcor as well? I didn't need to add cuBLAS to the link libraries - only cuFFT was needed.

piyushrpt:
@lijun99  might be able to answer better, but I rember cublas context being used in the .cu files. 
https://github.com/isce-framework/isce2/blob/da9a8afaf593e386cfe55a5495c59e2141481794/components/zerodop/GPUampcor/cuda/GPUamp.cu#L261

lijun99:
GPUamp.cu uses the cublas within device code, which is no longer supported by the recent CUDA. We'd better drop that. 
For PyCuAmpcor, I probably used a function from cublas which is no longer needed. I can clean it up (remove #include <cublas_v2.h> ....) so that only cufft is needed. 
I remember cublas in CMake is a little bit tricky, since nowadays CUDA installs cublas to system directories instead of /usr/local/cuda. I did something for pyre_cuda https://github.com/lijun99/pyre/blob/cuda/.cmake/pyre_cuda.cmake but don't remember the details. I can double check if needed.

piyushrpt:
Finally made the switch to cmake this weekend. This looks good. Only issue I can think of is that motif is not optional and using this on conda will cause failure on osx. We should probably make this optional since mdx doesn't build on conda for osx.

rtburns-jpl:
I think you'll be pleased to hear I've been using these scripts to build mdx on macos conda 😎 
If you don't have openmotif-dev and xorg-libxt though, it will silently fall back and not build mdx.
And if you have the cores to spare, make -j32 is lightning fast :)

piyushrpt:
Aha ... awesome. Then it must be the interaction between scons and openmotif (probably in finding the libraries) - thats the issue. 

I'm definitely a convert.

rtburns-jpl:
@piyushrpt any idea why circleci isn't reporting anymore? This passed both workflows and is ready.
PyCuAmpcor seems to be working, I tested it and got good results. Virginia has been using cmake for testing CPU ampcor scripts as well.

piyushrpt:
Things seem to be running - https://app.circleci.com/pipelines/github/isce-framework/isce2
Wondering if anything changed on their reporting

piyushrpt:
I also had to make the following changes on osx if I didnt want it installed under a local directory. You may have pointers already to fix this. Without these changes you get a @rpath linking error when running any of the workflows.

```
set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")

# add the automatically determined parts of the RPATH
# which point to directories outside the build tree to the install RPATH
set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)

# the RPATH to be used when installing, but only if it's not a system directory
list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
if("${isSystemDir}" STREQUAL "-1")
    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
endif("${isSystemDir}" STREQUAL "-1")
```

rtburns-jpl:
Have you tried setting DYLD_FALLBACK_LIBRARY_PATH? I didn't need to make any changes to rpath when testing locally

piyushrpt:
Have you been able to get DYLD_FALLBACK_LIBRARY_PATH to work on OS X Catalina as well? I think the issue might be related to the new OS. I seem to need those mac specific changes for cmake to build without RPATH issues.

rtburns-jpl:
Oh, I'm not on Catalina yet so I don't know. I'm fine to set up rpath then, and it looks like kitware has a good set of recommendations here:
https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#always-full-rpath
I think ideally we'd also use $ORIGIN so that the package is fully relocatable. I'll play around with this and see if I can get something working.

piyushrpt:
ha ha ha ... my set of instructions are from that page exactly. Will keep playing around. Seems like this is an issue on catalina only.

rtburns-jpl:
Unless there are any more issues/suggestions, I don't foresee any major changes here so I'll plan on merging this soon.

PR #139, opened 2020-05-29
cuDenseOffsets: expose --raw-osf and --corr-win-size option

yunjunz:
This PR exposes two more parameters for the `PyCuAmpcor/examples/cuDenseOffsets.py `: `--raw-over-samp-factor` to be able to change the oversampling factor of the raw data and `--corr-win-size` to be able to use a custom window. The default values are still the same as before.

It also improve the plotting in the `stripmapStack/MaskAndFilter.py`.

The summary of detailed changes are as below:

contrib/PyCuAmpcor/examples/cuDenseOffsets.py:
+ expose --raw-over-samp-factor option to change the raw data oversampling factor with default of 2
+ expose --corr-win-size option, which by default, is calculated automatically based on the search window size and raw oversampling factor
+ change the namespace key for the oversampling factor of correlation surface from 'oversample' to 'corr_oversample' to avoid ambiguity, the cmd option name is still the same, thus no different in the user side.
+ use vrt file for master / slave slc data loading folllowing https://github.com/isce-framework/isce2/pull/78
+ re-organize createParse() for readability
+ add example to argparse help

stripmapStack/MaskAndFilter.py:
+ fix a bug introduced in https://github.com/isce-framework/isce2/pull/112
+ add example usage to help message
+ mask pixels with SNR == 0, for plotting only without touching data files
+ show input parameter in the figure title
+ show percentage after SNR thresholding
+ assign nearest interpolation for matplotlib.pyplot.imshow() while plotting to avoid anti-aliasing

topsStack/grossOffsets: move obsolete basemap import from the top to the related commented code block

### DIFF
--- a/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
+++ b/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
@@ -107,6 +107,12 @@ def cmdLineParse(iargs = None):
     parser = createParser()
     inps =  parser.parse_args(args=iargs)
 
+    # check oversampled window size
+    if (inps.winwidth + 2 * inps.srcwidth) * inps.raw_oversample > 1024:
+        msg = 'input oversampled window size in the across/range direction '
+        msg += 'exceeds the current implementaion limit of 1024!'
+        raise ValueError(msg)
+
     return inps
 
 
--- a/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
+++ b/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
@@ -1,87 +1,107 @@
 #!/usr/bin/env python3
 
-# Author: Minyan Zhong, Lijun Zhu 
+# Author: Minyan Zhong, Lijun Zhu
 
-import argparse
+
 import os
+import argparse
+import numpy as np
+
 import isce
 import isceobj
 from isceobj.Util.decorators import use_api
-
-import numpy as np 
 from contrib.PyCuAmpcor.PyCuAmpcor import PyCuAmpcor
+
+
+EXAMPLE = '''example
+  cuDenseOffsets.py -m ./merged/SLC/20151120/20151120.slc.full -s ./merged/SLC/20151214/20151214.slc.full
+      --masterxml ./master/IW1.xml --outprefix ./merged/offsets/20151120_20151214/offset
+      --ww 256 --wh 256 --oo 32 --kw 300 --kh 100 --nwac 100 --nwdc 1 --sw 8 --sh 8 --gpuid 2
+'''
+
 
 def createParser():
     '''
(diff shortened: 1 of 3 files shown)


hfattahi (approved):

piyushrpt:
@lijun99 - can you take a look at these proposed changes?

lijun99:
The PyCuAmpcor related part looks good to me. But one caveat is that the current implementation limits the oversampled window size max(width, height)<=1,024 (the max threads per block for CUDA). It's better to add a check, so that raw-over-samp-factor*(windowsize+searchrange), or the oversampled slave chip size <=1024.

That limit comes from a routine for prefix sum. I can change/add for larger window support if it is needed. But the algorithm probably will be less efficient. Let me know.

yunjunz:
@lijun99 A check on oversampled window size sounds good. I sometimes got a memory error while playing the parameters ;) 

Could you write the equation for this oversampled window size using the name of the arguments explicitly? I am unclear about the following equation. Should the corr-over-sample-factor be included as well?

> raw-over-samp-factor*(windowsize+searchrange)

lijun99:
(windowSizeWidth+2* haldSearchRangeAcross)* rawDataOversamplingFactor <=1024. 
Only the across direction is enforced; the down direction can be over 1024. Of course, some modifications can be made if support for larger window size is needed. 

I remember rawDataOversamplingFactor is in general taken as 2 for antialiasing. Is there a reason for the factor being other than 2? Just curious.

yunjunz:
Thank you @lijun99. 

From my testing with different configurations, a larger rawDataOversamplingFactor gives a higher sensitivity, similar to the correlation oversampling factor. So I thought it will be useful to let it exposed to users to play with.

vbrancat:
The overall changes look good to me. Just a minor comment that popped up with using PyCuAmpcor lately. It would be good if the user can choose how the cross-correlation is computed (spatial or frequency domain). This functionality is somehow hidden in the code and I think it should be given as a choice. Computing the cross-correlation in the spatial or frequency domain makes a difference for optical images :). For the rest, great job!

vbrancat (approved):

vbrancat (approved):

vbrancat:
The overall changes look good to me. Just a minor comment that popped up with using PyCuAmpcor lately. It would be good if the user can choose how the cross-correlation is computed (spatial or frequency domain). This functionality is somehow hidden in the code and I think it should be given as a choice. Computing the cross-correlation in the spatial or frequency domain makes a difference for optical images :). For the rest, great job!

rtburns-jpl (approved):
LGTM! @vbrancat, I have a simple patch I've been using that exposes the algorithm option, so I can add that at a later point if needed.

yunjunz:
Could anyone merge this PR?

Issue #137, opened 2020-05-26
Baseline grid bug


sssangha:
I have documented a bug whereby there is an abrupt sign-change when generating perpendicular baseline grids. I have generated these products using the "baselineGrid.py" script packaged through stackSentinel, which is essentially a wrapper of this ISCE script "components/zerodop/baseline/Baseline.py". I have attached "baselineGrid.py" as a reference here within a zip file.

The master scene used was "S1B_IW_SLC__1SDV_20181210T000734_20181210T000752_013972_019ECE_FB9D.zip", and the slave scenes used were "S1B_IW_SLC__1SDV_20171203T000703_20171203T000730_008547_00F2A8_F803.zip, S1B_IW_SLC__1SDV_20171203T000727_20171203T000745_008547_00F2A8_DCE6.zip"

Within this zip file the extracted baseline grids may also be found, but I will illustrate the problem below:
<img width="402" alt="Screen Shot 2020-05-26 at 12 49 49 PM" src="https://user-images.githubusercontent.com/13227405/82944064-6beabc80-9f4f-11ea-8903-f1707e454d4d.png">

I made some print statements to track changes in the variables involved in the estimate of the direction (i.e. "direction = np.sign(np.dot( np.cross(targxyz-mxyz, sxyz-mxyz), mvel))"), and here is an example of two sets of variables corresponding to adjacent pixels along range for which there was a sign flip:

> 

############################################################################################ 
direction -1.0 
targxyz [  13843.73542219 5020808.95559147 3920247.75736679] 
mxyz [-431237.83432596 5613389.08465543 4277645.57748178] 
sxyz [-431234.7701128  5613379.85014268 4277646.91694975] 
mvel [ 1404.18442248  4599.24232307 -5877.04480614] 
Bperp[jj] -6.0630693 ############################################################################################ 
direction 1.0 
targxyz [  26673.98226566 5019284.63081567 3922120.34709414] 
mxyz [-431237.83432596 5613389.08465543 4277645.57748178] 
sxyz [-431234.80521498 5613379.73516128 4277647.06387625] 
mvel [ 1404.18442248  4599.24232307 -5877.04480614] 
Bperp[jj] 6.2536173

> 

After looking carefully across this sign-flip in the field, it was my impression that the field would be smoothly varying. I.e. if I wrote out the absolute value of all pixels in the field, there would no longer be an abrupt jump in the field. Thus, I believe there could be an issue in the baseline equation, but I haven't been able to pinpoint a bug.
[...]

PR #141, opened 2020-06-04
Baseline bug fix

sssangha:
Following solution suggested by @piyushrpt in issue #137, estimation of slave satellite position adjusted in topsApp and zerodop associated baseline codes to address artifacts in metadata fields.


dbekaert:
@sssangha believe this fixes it for both topsApp and StackSentinel.

piyushrpt (approved):

piyushrpt:
It would be good to open an issue to address this for stripmap stack as well

PR #156, opened 2020-07-04
Fixing linking order and flags for mdx

### LINKED ISSUE #155: mdx from cmake pipeline results in segfault with conda compilers
I was looking to replacing scons with cmake for conda recipes. Note that there we no compilation failures.

1. scons and cmake are using different sources. CMakeLists.txt should be updated to use mdx_subs.F instead of mdx.F. Even with that change, there are some issues.
2. Looks like cmake generated mdx results in segfaults in calls to "mdxsub" 
   - All the files are being compiled straight into an executable here
   - Tried replacing the conda compilers with system compilers but the problem persisted
   - I suspect the culprit might be in one of the three
         - there were always bugs in mdx and conda's optimization flags causes the error
         - sequence of compilation/ linking somehow matters for this case
         - additional args might be needed for linking in one go
3. scons generated mdx seems to work
   - Files are compiled into a static lib first and then compiled into an executable 

I tried for a fair bit to track down the issue by matching compiler and linking flags but that didnt make a difference. Somehow generating a static lib and then using to generate the executable seems to produce this behavior. Attaching the logs here - in case something stands out immediately that I missed. 
[scons.log](https://github.com/isce-framework/isce2/files/4871911/scons.log)
[cmake.log](https://github.com/isce-framework/isce2/files/4871912/cmake.log)

piyushrpt:
Fixes #155 

1. disable openmp flag explicitly for mdx 
2. linking order of libraries - I would like to use the nice target management but that seems to make a mess of the order in which mdx expects libs. 

This PR should probably be updated with changes to TargetX11 and TargetMotif. Feel free to push to this branch and make changes.

### DIFF
--- a/.cmake/TargetX11.cmake
+++ b/.cmake/TargetX11.cmake
@@ -5,9 +5,6 @@ set(components
 
 find_package(X11 COMPONENTS ${components})
 
-# The X11::{component} targets only exist for CMake ≥ 3.14,
-# so we create them here for backwards compatibility.
-
 if(X11_FOUND)
 
     # make X11 look like a regular find_package component
--- a/contrib/mdx/CMakeLists.txt
+++ b/contrib/mdx/CMakeLists.txt
@@ -3,22 +3,30 @@ if(TARGET Motif::Motif
    AND TARGET X11::Xau
    AND TARGET X11::Xt
    )
+
+    #Only for this folder - ensure openmp is not used for compiling
+    #Modifying flags here have directory scope
+    if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "GNU")
+        string(REPLACE "-fopenmp" " " CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS}")
+    endif()
+    
     add_executable(mdx
         src/graphx_mdx.c
         src/rdf_reader_subs.f
         src/mdx_main.F
-        src/mdx.F
+        src/mdx_subs.F
         )
     target_compile_definitions(mdx PRIVATE SUN IO64)
-    if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "GNU")
-        target_compile_definitions(mdx PRIVATE GFORTRAN)
-    endif()
+    target_include_directories(mdx PUBLIC 
+        ${X11_Xt_INCLUDE_PATH}
+        ${X11_X11_INCLUDE_PATH}
+        ${MOTIF_INCLUDE_DIR})
     target_link_libraries(mdx PUBLIC
-        Motif::Motif
-        X11::X11
-        X11::Xau
-        X11::Xt
+        ${X11_X11_LIB}
+        ${MOTIF_LIBRARIES}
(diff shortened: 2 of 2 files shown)


piyushrpt:
Is there a better solution? The solution here does not change affect "X11::" targets at all. It explicitly sets up "X11isce::" and uses it. So, this will not get in the way of anything else expecting "X11::"

rtburns-jpl:
I like that you gave your specialized targets different names but I'd like the X11:: targets to still exist for use in other targets if needed in the future. Since we only need them for mdx, why not something like this inside contrib/mdx/CMakeLists.txt?
```cmake
add_library(isce2::motif IMPORTED INTERFACE)
target_link_libraries(isce2::motif INTERFACE
        X11::X11
        Motif::Motif
        X11::Xt)
target_link_libraries(mdx PRIVATE isce2::motif)
``` 

But taking a step back, I still feel there is a still undiagnosed underlying issue behind this and the CMAKE_Fortran_FLAGS being set improperly. I'd like to see if there is a more idiomatic solution for this but still haven't been able to reproduce your errors. Do you have a dockerfile or more details about your environment I can use to recreate this?

piyushrpt:
The Fortran Flags is a conda issue (where it is set by default) - their argument, why are you using fortran compilers in something meant for C/C++/Python.  I think the solution above runs into same issue, the order wont be preserved in new cmake - due to interface_libraries being setup in FindX11. 

Another option is to explicitly using X11_{COMPONENT}_INCLUDE_PATH and X11_{COMPONENT}_LIB directly without using the interfaces like @lijun99 mentioned.

rtburns-jpl:
Yeah, are the fortran flags being set automatically in a `conda build`? I suspect this could be due to an environment variable and would like to repro this. Knowing what distro, packages, etc you're using would be a big help.

I'd be fine using the explicit FindX11 variables if we can't use the targets. But as far as I can tell the target_link_libraries command should preserve order, whether the arguments are library names or targets.

piyushrpt:
conda build sets env variable FFLAGS that includes -fopenmp.  
The order of libraries need to be necessarily preserved if the interfaces themselves define other interface_link_libraries

https://github.com/Kitware/CMake/blob/78df084c7abc9dfa2561b16f9d319c417a3cf0d5/Modules/FindX11.cmake#L724-L730

Then cmake's internal algorithm kicks in to try and resolve the linking order - which is fairly complicated from couple of blog posts that i came across. 

Sounds like, using the FindX11 variables might be a good compromise that works across different versions of cmake.

rtburns-jpl:
Agreed. That's weird to me that conda-forge sets all those variables, but I guess we're the weird case that simply enabling openmp breaks our build. I'd probably prefer that the -fopenmp is patched out in the recipe build.sh, but I'm fine with doing it here instead if that's more convenient.

I think you're right about the FindX11 variables. I just pushed a commit to motif-deps - let me know if that enforces linking order properly.

rtburns-jpl (approved):
Looks great! Thanks for taking the time to track this down

PR #157, opened 2020-07-06
Add argument mismatch workaround for gfortran 10

rtburns-jpl:
GCC 10 seems to tighten up some restrictions on fortran implicit casts. Was getting errors similar to the following:
```
/home/user/work/isce2/components/mroipac/ampcor/src/ampcor.F:503:50:

  483 |                     call getLineBand(imgAccessor1, c_refimg(:,i_yy), i_xx, i_lineno)
      |                                                   2
......
  503 |                    call getLineBand(imgAccessor1, r_refimg(:,i_yy), i_xx, i_lineno)
      |                                                  1
Error: Type mismatch between actual argument at (1) and actual argument at (2) (REAL(4)/COMPLEX(4)).
```
I think this is something we should look into and fix, but this workaround lets it compile for the time being. Perhaps someone with more fortran experience knows a better approach.

### DIFF
--- a/.cmake/isce2_buildflags.cmake
+++ b/.cmake/isce2_buildflags.cmake
@@ -8,6 +8,11 @@ add_compile_options(
     $<$<COMPILE_LANGUAGE:Fortran>:-ffixed-line-length-none>
     $<$<COMPILE_LANGUAGE:Fortran>:-fno-range-check>
     $<$<COMPILE_LANGUAGE:Fortran>:-fno-second-underscore>)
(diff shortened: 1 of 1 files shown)


piyushrpt:
mdx has a whole bunch of similar code where int and pointers are interchangably used. This looks good if mdx works with these changes.

rtburns-jpl:
Yep, can confirm this is working for me in mdx and everything else I've tried using. Arch and fedora are already on gcc 10 and looks like debian sid is soon to follow.

piyushrpt (approved):

PR #158, opened 2020-07-09
Fix violations of ODR

rtburns-jpl:
I've been getting linker errors such as
```
/usr/bin/ld: CMakeFiles/alos.dir/src/ALOS_pre_process/read_ALOSE_data.c.o:(.bss+0x40f0): multiple definition of `debug'; CMakeFiles/alos.dir/bindings/alosmodule.cpp.o:(.bss+0x30): first defined here
```
and assuming they were due to my improper configuration. But upon a closer look the ALOS helper libs have clear violations of the ODR and are therefore [IFNDR](https://quuxplusone.github.io/blog/2019/08/02/the-tough-guide-to-cpp-acronyms/#ifndr).

The updated extension modules import properly without errors but I'd appreciate if someone who uses them could run a quick sanity check.


rtburns-jpl:
Will revisit pending https://github.com/isce-framework/isce2/issues/167

rtburns-jpl:
-fno-common is now the default in GCC 10 and Clang 11, so we should make sure these are fixed - see https://github.com/conda-forge/isce2-feedstock/pull/31. I also added some cmake logic to use -fno-common on older compilers when supported, which should help to catch these early and prevent regressions.

piyushrpt (approved):

piyushrpt:
This code originally originated in UCSD and this code structure is very common in GMTSAR. It might be a good idea to give them a heads up.

rtburns-jpl:
Ah, you're right. For example https://github.com/gmtsar/gmtsar/blob/master/gmtsar/gmtsar.h is quite similar. I'll try to compile gmtsar with -fno-common and report any issues I find.

PR #160, opened 2020-07-16
Use openmp for resamp_slc

piyushrpt:
Noticed slow performance for topsApp when using cmake. 

Related note:
Use -DOpenMP_Fortran_LIB_NAMES="omp"  when using clang. Uses gomp by default.

PR #162, opened 2020-07-18
Add openmp to stdproc modules

piyushrpt:
Adding openmp to stdproc modules.
crossmul and resamp_slc are used in active workflows and others possibly in historic/ unmaintained workflows.


piyushrpt:
Same as #160 but rebased on main branch

rtburns-jpl (approved):

PR #163, opened 2020-07-22
Bake in autoRIFT and add CMake support

rtburns-jpl:
Add autoRIFT to contrib/ for simplified builds + redistribution
Regards https://github.com/conda-forge/isce2-feedstock/pull/19

The README is really nice looking and pleasant to read so I kept it in the subdirectory. To keep the repo size down though, I removed the .png figures and changed the README \ links to point to an external repo.
Preview at https://github.com/rtburns-jpl/isce2/blob/bake-in-autorift/contrib/geo_autoRIFT/README.md

Tested build locally with macOS conda and Nix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,7 @@ jobs:
             pwd
             mkdir config build install
             . /opt/conda/bin/activate root
-            conda install --yes cython gdal h5py libgdal pytest numpy fftw scipy scons hdf4 hdf5 netcdf4 libgcc libstdcxx-ng cmake astropy
+            conda install --yes cython gdal h5py libgdal pytest numpy fftw scipy scons hdf4 hdf5 netcdf4 libgcc libstdcxx-ng cmake astropy opencv
             yum install -y uuid-devel x11-devel motif-devel jq gcc-gfortran
             ln -s /opt/conda/bin/cython /opt/conda/bin/cython3
             cd /opt/conda/lib
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,8 @@ find_package(Python 3.5 REQUIRED COMPONENTS Interpreter Development
 find_package(FFTW REQUIRED)
 find_package(Motif)
 find_package(OpenMP REQUIRED COMPONENTS C CXX Fortran)
+find_package(OpenCV COMPONENTS core highgui imgproc)
+
 # Find these, and create IMPORTED INTERFACE libraries for them if they exist
 include(TargetGDAL)
 include(TargetMotif)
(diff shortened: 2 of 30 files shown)


piyushrpt:
Do we want to pull in geoGrid and autoRIFT physically or as a submodule possibly. We may have to actively migrate any changes made to that repo. What do you prefer?

rtburns-jpl:
Well, if it's a submodule, users will have to use `--recursive` or some lesser-known submodule commands to get the full repo. Also it looks like we have to dump a subdir rather than the toplevel folder, so the autoRIFT repo would have to be restructured.

If you're open to making updates here directly, it can simplify versioning to have everything in one place, and might make your life easier. But if you don't want to, I'm fine with regularly pulling your releases.

piyushrpt:
autoRIFT has a user base that works with optical imagery as well and doesn't need ISCE for it. It has its own conda feedstock as well - to cater to those users. There are a couple of production pipelines that are using isce+autorift at ASF and they rely on the conda packages for it. I think asking those developers to make changes in 2 places might be cumbersome.

rtburns-jpl:
Ok, thanks. I guess monorepo approach and submodule are both off the table then. I'm fine with regularly updating the code to align with autoRIFT releases, since I think that would greatly simplify our builds. But if there are benefits to "baking in" during package construction instead, perhaps I should contribute these CMakeLists upstream?

piyushrpt:
One option is to add CMakeLists to autoRIFT repo and use same approach as scons - if it sees the corresponding sub-folders, add that in parent CMakeLists. 

@leiyangleon - thoughts on pull in frequently from autoRIFT. Are you willing to let us know when you push a change to autoRIFT?

leiyangleon:
> One option is to add CMakeLists to autoRIFT repo and use same approach as scons - if it sees the corresponding sub-folders, add that in parent CMakeLists.
> 
> @leiyangleon - thoughts on pull in frequently from autoRIFT. Are you willing to let us know when you push a change to autoRIFT?

Possibly by the end of August. Alex and I are still working on some new features of it, which has been a while. We will shoot for issuing the next release before Sep.

rtburns-jpl:
@leiyangleon Great! I see that you've been tagging and releasing regularly via github, at https://github.com/leiyangleon/autoRIFT/releases. Assuming you're planning to stick with that approach, I'll subscribe to releases on your repo. Then I'll automatically get a heads up whenever there's a new version for me to bake in.

leiyangleon:
> @leiyangleon Great! I see that you've been tagging and releasing regularly via github, at https://github.com/leiyangleon/autoRIFT/releases. Assuming you're planning to stick with that approach, I'll subscribe to releases on your repo. Then I'll automatically get a heads up whenever there's a new version for me to bake in.

@rtburns-jpl Sure, I will stick with that approach. So feel free to subscribe on my repo.

piyushrpt (approved):

PR #166, opened 2020-07-27
Add SAOCOM sensor

asolarte:
Hello, I leave for consideration the files to incorporate SAOCOM data processing.

### DIFF
--- a/components/isceobj/Sensor/SAOCOM_SLC.py
+++ b/components/isceobj/Sensor/SAOCOM_SLC.py
@@ -1,0 +1,450 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# copyright: 2010 to the present, california institute of technology.
+# all rights reserved. united states government sponsorship acknowledged.
+# any commercial use must be negotiated with the office of technology transfer
+# at the california institute of technology.
+# 
+# this software may be subject to u.s. export control laws. by accepting this
+# software, the user agrees to comply with all applicable u.s. export laws and
+# regulations. user has the responsibility to obtain export licenses,  or other
+# export authority as may be required before exporting such information to
+# foreign countries or providing access to foreign persons.
+# 
+# installation and use of this software is restricted by a license agreement
+# between the licensee and the california institute of technology. it is the
+# user's responsibility to abide by the terms of the license agreement.
+#
+# Author: Andrés Solarte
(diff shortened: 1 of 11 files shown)


examples/input_files/reference_SAOCOM.xml
```diff
@@ -0,0 +1,46 @@
...
+    <property name="XEMTFILE">
+        S1A_OPER_SAR_EOSSP__CORE_L1A_OLF_20200610T203550.xemt
+    </property>
+    <property name="XMLFILE">
+        S1A_OPER_SAR_EOSSP__CORE_L1A_OLF_20200610T203550/Data/slc-acqId0000076824-a-sm8-0000000000-s8dp-vv.xml
+    </property>
+    <property name="OUTPUT">
+        slc-acqid0000076824-a-sm8-s8dp-vv.slc
+    </property>
+    <property name="unwrapper name">snaphu</property>
```
piyushrpt, line 14: This unwrapper name entry here is probably a typo.

components/isceobj/Sensor/SAOCOM_SLC.py
```diff
@@ -0,0 +1,449 @@
...
+        
+        self.dopplerRangeTime = self._xmlFileParser.dopRngTime
+        self.rangeRefTime = self._xmlFileParser.trg
+        self.rangeFirstTime = self._xmlFileParser.rangeStartTime
+        
+    def extractImage(self):
+        """
+        Exports GeoTiff to ISCE format.
+        """
+        import gdal
```
piyushrpt, line 245: I would change this to 
```
from osgeo import gdal
```

piyushrpt:
@asolarte - thank you so much for this contribution. Just to understand - this is for the stripmap only correct? 

Also, to let you know - if you add an unpackSAOCOM.py script to contrib/stack/stripmapStack folder - you will be able to generate stacks of coregistered SLCs as well with the stack processor.

examples/input_files/reference_SAOCOM.xml
asolarte: Hi, I left that tag, since it works better for me than the unwrapper that runs by default.

components/isceobj/Sensor/SAOCOM_SLC.py
asolarte: > @asolarte - thank you so much for this contribution. Just to understand - this is for the stripmap only correct?
> 
> Also, to let you know - if you add an unpackSAOCOM.py script to contrib/stack/stripmapStack folder - you will be able to generate stacks of coregistered SLCs as well with the stack processor.

@piyushrpt  - Yes, It's for stripmap only.
Thank you so much... I'll work on the unpackSAOCOM.py script

piyushrpt (approved):

hfattahi (approved):

PR #168, opened 2020-07-31
Fix python extension linking

### LINKED ISSUE #167: Link with DataAccessor_static for stdproc components with CMake
Just like openmp, I think link DataAccessor_static linking is missing for stdproc components.

This seems to work fine on OSX but fails on Linux - as reported by @hfattahi

rtburns-jpl:
Diff is on the larger side due to some sed usage - primary changes are:
- CMake minimum version is bumped to 3.13 to use relative paths in targets
- Use `isce2_add_staticlib` and `isce2_add_cdll` helpers for creating python extensions, which closely mimic the SCons behavior for these targets.
- Set up `Python_add_library` and `isce2_add_cdll` to create trivial python-import tests, verifying that there are no missing link libraries
- Static libs are given properly namespaced names, and use more robust `::` alias targets when linking into other targets

In general this should improve robustness, ease usage alongside other cmake projects, and prevent similar errors in the future.

Fixes https://github.com/isce-framework/isce2/issues/167


rtburns-jpl:
Ah, looks like TARGET_FILE_BASE_NAME requires CMake 3.15. Will look into a suitably compatible replacement

piyushrpt (approved):

piyushrpt:
Do you want to rebase this on the new CI stuff for testing?

rtburns-jpl:
Rebased cleanly and passed all tests locally. Let's see how the CI likes it.

PR #173, opened 2020-08-03
Add Numpy + OpenCV autodetection to SCons

rtburns-jpl:
@EJFielding @leiyangleon

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -110,7 +110,7 @@ if confinst.CheckCHeader("numpy/arrayobject.h"):
 else:
     print("Numpy not found.")
 
-    if sys.version_info.major == 2:
+    if sys.version_info[0] == 2:
         raise RuntimeError("Cannot autodetect numpy from python2")
 
     print("Attempting to autodetect...")
--- a/SConstruct
+++ b/SConstruct
@@ -103,6 +103,28 @@ inst = env['PRJ_SCONS_INSTALL']
 ####new part
 #####PSA. Check for header files and libraries up front
 confinst = Configure(env)
+
+# Try to autodetect numpy
+if confinst.CheckCHeader("numpy/arrayobject.h"):
+    print("Numpy header found.")
+else:
+    print("Numpy not found. Attempting to autodetect...")
+    try:
+        import numpy
+        confinst.env["CPPPATH"].append(numpy.get_include())
+        print("Added autodetected numpy include to CPPPATH.")
+    except ImportError:
+        raise RuntimeError("""Could not autodetect numpy include.
+            Please add the numpy include dir to your CPPPATH.""")
+
+# Check for OpenCV
+if confinst.CheckCXXHeader("opencv2/imgproc/imgproc.hpp"):
+    confinst.env["HAVE_OPENCV"] = True
+    print("OpenCV detected - enabled.")
+else:
+    confinst.env["HAVE_OPENCV"] = False
+    print("OpenCV not detected - disabled.")
+
 hdrparams = [('python3 header', 'Python.h', 'Install python3-dev or add path to Python.h to CPPPATH'),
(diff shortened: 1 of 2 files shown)


SConstruct
```diff
@@ -103,6 +103,28 @@ inst = env['PRJ_SCONS_INSTALL']
...
 #####PSA. Check for header files and libraries up front
 confinst = Configure(env)
+
+# Try to autodetect numpy
+if confinst.CheckCHeader("numpy/arrayobject.h"):
+    print("Numpy header found.")
+else:
+    print("Numpy not found. Attempting to autodetect...")
+    try:
+        import numpy
```
piyushrpt, line 113: This is tricky because scons could be using a different python version than the one being used for installation - notably macports. Its lot easier with cmake to do this.

EJFielding: Yes, I think I have always used the scons from Python2. Not sure why I got started with that, but I never changed it.

rtburns-jpl: True. Unless you know how to have scons invoke the python used for installation, maybe I should just disable this when scons is using python 2.

rtburns-jpl:
Hmm, I removed opencv from from the installation prereq list to test the autodetect fallback, but dependency resolution still failed. Is there another package conflict occurring in our prereqs?

piyushrpt:
hdf5 might need to be pinned. I think the new version still hasnt propagated to all the feedstocks.

EJFielding:
We also had a problem with HDF5 library requirement. @nicodalse had installed the ISCE2 requirements with conda and did a build. The resulting executables were pulling in an HDF5 library from /lib64 on our Linux server instead of using the one from his conda installation. I don't know if this is the same problem. I recommended that he try using `cmake` instead of `scons` but he had other problems.

nicodalse:
Yes, opencv gave me a lot of conflicts and the hdf5 library get me this error also in a new virtual env:

```
SafetyError: The package for hdf5 located at /u/aa/seno/miniconda3/pkgs/hdf5-1.10.5-nompi_h3c11f04_1104
appears to be corrupted. The path 'lib/libhdf5_cpp.so.103.1.0'
has an incorrect size.
 reported size: 593616 bytes
 actual size: 347944 bytes
SafetyError: The package for hdf5 located at /u/aa/seno/miniconda3/pkgs/hdf5-1.10.5-nompi_h3c11f04_1104
appears to be corrupted. The path 'lib/libhdf5_fortran.so.102.0.0'
has an incorrect size.
 reported size: 405384 bytes
 actual size: 299176 bytes
SafetyError: The package for hdf5 located at /u/aa/seno/miniconda3/pkgs/hdf5-1.10.5-nompi_h3c11f04_1104
appears to be corrupted. The path 'lib/libhdf5_hl.so.100.1.2'
has an incorrect size.
 reported size: 163864 bytes
 actual size: 144968 bytes
SafetyError: The package for hdf5 located at /u/aa/seno/miniconda3/pkgs/hdf5-1.10.5-nompi_h3c11f04_1104
appears to be corrupted. The path 'lib/libhdf5_hl_cpp.so.100.1.3'
has an incorrect size.
 reported size: 23512 bytes
 actual size: 11336 bytes
SafetyError: The package for hdf5 located at /u/aa/seno/miniconda3/pkgs/hdf5-1.10.5-nompi_h3c11f04_1104
appears to be corrupted. The path 'lib/libhdf5hl_fortran.so.100.0.4'
has an incorrect size.
 reported size: 168008 bytes
 actual size: 99208 bytes
```

piyushrpt:
these errors seem to be driver driven due to environment issues and not due to conda. The conda issue is that none of the package versions are pinned and there should be no issues if they are pinned like it is done on raider / mintpy. Quickest way is to see dependencies used by isce2 conda package and use same package versions for hdf5 and gdal.

EJFielding:
I heard from @nicodalse that he was able to merge my `uavsar` PR branch with the `cmake` branch from @rtburns-jpl and get the build with `cmake` to work in his Linux conda environment.

PR #180, opened 2020-08-19
CMake: Fix missing modules required for UAVSAR stripmap stack

rtburns-jpl:
Thanks to @nicodalse for pointing out these missing modules, which were needed to run the UAVSAR stripmap stack.

This is based on https://github.com/isce-framework/isce2/pull/168 - will rebase and remove draft status once that is merged.


hfattahi:
@rtburns-jpl Would you please rebase and merge if this is ready.

PR #186, opened 2020-08-26
Fixed namespace bug in AkimaLib.cpp

sharkbig:
add using namespace bug


rtburns-jpl (approved):
Thanks for contributing!

PR #187, opened 2020-08-29
topsStack/FilterAndCoh: use fine.cor(.full) for multilooked (full-reso) cpx cor file

yunjunz:
This PR adds i) `.full` suffix to the output complex coherence filename only if input multilook number is 1 and ii) skip running `Looks()` when the multilook number is 1.

Here are the detailed changes

+ topsStack/FilterAndCoherence.py:
  - add .full suffix to the complex coherence filename only if the multilook number is 1
  - for multilook number as 1, do not run looks sub-module

+ topsStack/Stack.filter_coherence(): add .full suffix only when `azLooks * rgLooks == 1`

+ topsStack/extractCommonValidRegion: check stack/*.xml file existance


yunjunz:
I forgot to make a feature branch, will close this one and re-post.

Issue #126, opened 2020-05-02
Update system calls to use gdal2isce_xml python interface in next release


piyushrpt:
Addresses issue in #125 
Target - to be fixed just before next release

PR #188, opened 2020-08-29
topsStack: use OMP_NUM_THREADS for topo and drop ".full" for non-full-reso cpx cor file in FilterAndCoh

yunjunz:
This PR adds a few changes to the topsStack processor:

+ topsStack/topo: use `OMP_NUM_THREADS` if it's defined in the environment variables for the parallel processing; otherwise, use all available CPUs, same as before.
+ topsStack/FilterAndCoherence: drop the `.full` suffix to the output complex coherence filename if multilooking (>1) is applied.
+ topsStack/FilterAndCoherence: skip running `Looks()` module if no multilooking is applied.
+ topsStack/extractCommonValidRegion: check stack/*.xml file existance
+ replace the os.system() call of `gdal2isce_xml.py` with python funcation call, in `stripmapStack/topo.py` and `applications/downsampleDEM.py` (#126).

Supersedes #187.

### DIFF
--- a/contrib/stack/topsStack/FilterAndCoherence.py
+++ b/contrib/stack/topsStack/FilterAndCoherence.py
@@ -128,25 +128,31 @@ def estCpxCoherence(slc1_file, slc2_file, cpx_coh_file, alks=3, rlks=9):
     from isceobj.TopsProc.runBurstIfg import computeCoherence
     from mroipac.looks.Looks import Looks
 
+    # get the full resolution file name
+    if alks * rlks == 1:
+        cpx_coh_file_full = cpx_coh_file
+    else:
+        cpx_coh_file_full = cpx_coh_file+'.full'
+
     # calculate complex coherence in full resolution
-    computeCoherence(slc1_file, slc2_file, cpx_coh_file)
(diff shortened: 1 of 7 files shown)


yunjunz:
Could someone take a look at this PR?

contrib/stack/topsStack/FilterAndCoherence.py
```diff
@@ -68,7 +68,7 @@ def runFilter(infile, outfile, filterStrength):
 def runFilter_gaussian(infile, outfile, filterStrength):
     from isceobj import Filter
     
-    logger.info("Applying power-spectral filter")
+    logger.info("Applying gaussian filter")
```
hfattahi, line 71: The filter is not Gaussian. It is the power-spectral adoptive filter.

hfattahi (approved):
LGTM. Thanks @yunjunz. The filter applied is not Gaussian as I left a comment below.

contrib/stack/topsStack/FilterAndCoherence.py
yunjunz: Thank you @hfattahi for the clarification, it's now rolled back to the original accurate msg.

piyushrpt (approved):

PR #190, opened 2020-09-10
make sure the orbit file glob also finds files starting with S1B

vincentschut:
Orbit files from Sentinel1B start with "S1B". However, the file globbing in the sentinel1.py code had "S1A" hardcoded, and thus it would fail to find orbit files for S1B.


piyushrpt (approved):

rtburns-jpl (approved):
Looks good. Thanks!

PR #192, opened 2020-10-07
Sentinel1 GRD thermal noise removal

vincentschut:
This PR implements the Thermal Noise Removal for Sentinel-1 GRD data in the rtcApp. It finalizes what was already there, and further enhances it (azimuth noise vectors, more accurate interpolation). A new parameter is added to rtcApp: "apply thermal noise correction", which defaults to False to not change the default behaviour.

The implementation tries to follow (the spirit of) https://sentinel.esa.int/documents/247904/2142675/Thermal-Denoising-of-Products-Generated-by-Sentinel-1-IPF. (though it differs a bit implementation-wise from what is suggested in that document, to improve performance).

I've tested this on several S1 GRD products with and without the noise correction applied, and then subtracted both results to check the final noise image. The noise images look as expected.

I hope this is a welcome addition, and that you are willing to consider inclusion in ISCE2.

PS sorry for the whitespace changes that kind of disturb the diff view, it's my editor doing that automatically. If it bothers you I could create a new PR without those.

### DIFF
--- a/components/isceobj/RtcProc/runPreprocessor.py
+++ b/components/isceobj/RtcProc/runPreprocessor.py
@@ -42,7 +42,7 @@ def runPreprocessor(self):
         frame.product.endingSlantRange = r0max
 
         try:
-            reference = extract_slc(frame, slantRange=(not slantRangeExtracted))
+            reference = extract_slc(frame, slantRange=(not slantRangeExtracted), removeNoise=self.apply_thermal_noise_correction)
             success=True
             if not slantRangeExtracted:
                 r0min = frame.product.startingSlantRange
@@ -72,10 +72,10 @@ def runPreprocessor(self):
     catalog.printToLog(logger, "runPreprocessor")
     self._grd.procDoc.addAllFromCatalog(catalog)
 
-def extract_slc(sensor, slantRange=False):
+def extract_slc(sensor, slantRange=False, removeNoise=False):
 #    sensor.configure()
     sensor.parse()
-    sensor.extractImage()
+    sensor.extractImage(removeNoise=removeNoise)
    
     if slantRange:
         sensor.extractSlantRange()
--- a/components/isceobj/RtcProc/runPreprocessor.py
+++ b/components/isceobj/RtcProc/runPreprocessor.py
@@ -7,6 +7,7 @@ import logging
 import isceobj
 import copy
 import os
+import inspect
 logger = logging.getLogger('isce.grdsar.runPreprocessor')
 
 def runPreprocessor(self):
@@ -75,7 +76,12 @@ def runPreprocessor(self):
 def extract_slc(sensor, slantRange=False, removeNoise=False):
 #    sensor.configure()
     sensor.parse()
-    sensor.extractImage(removeNoise=removeNoise)
+    sensor_extractImage_spec = inspect.getfullargspec(sensor.extractImage)
+    if "removeNoise" in sensor_extractImage_spec.args or "removeNoise" in sensor_extractImage_spec.kwonlyargs:
+        sensor.extractImage(removeNoise=removeNoise)
+    else:
+        print('Noise removal requested, but sensor does not support noise removal.')
+        sensor.extractImage()
    
     if slantRange:
         sensor.extractSlantRange()
--- a/README.md
+++ b/README.md
@@ -681,6 +681,7 @@ The inputs are Sentinel GRD zipfiles
         <property name="epsg id">32618</property>
         <property name="geocode spacing">100</property>
         <property name="geocode interpolation method">bilinear</property>
+        <property name="apply thermal noise correction">True</property>
         <component name="reference">
         <property name="safe">$dir$/rtcApp/data/S1A_IW_GRDH_1SDV_20181221T225104_20181221T225129_025130_02C664_B46C.zip</property>
         <property name="orbit directory">$dir$/orbits</property>
--- a/applications/rtcApp.py
+++ b/applications/rtcApp.py
@@ -192,6 +192,14 @@ GEOCODE_LIST = Application.Parameter(
      doc = "List of products to geocode."
                                       )
 
+APPLY_THERMAL_NOISE_CORRECTION = Application.Parameter(
+        'apply_thermal_noise_correction',
+        public_name='apply thermal noise correction',
+        default=False,
+        type=bool,
+        mandatory=False,
+        doc = 'Flag to apply thermal noise correction. Currently only available for Sentinel-1.')
+
 
 #Facility declarations
 REFERENCE = Application.Facility(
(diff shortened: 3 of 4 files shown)


hfattahi (approved):
Thank you very much @vincentschut. This is a great addition. Would you please attach an image of the computed noise for records (Perhaps before noise correction, after and the difference).

components/isceobj/RtcProc/runPreprocessor.py
```diff
@@ -72,10 +72,10 @@ def runPreprocessor(self):
     catalog.printToLog(logger, "runPreprocessor")
     self._grd.procDoc.addAllFromCatalog(catalog)
 
-def extract_slc(sensor, slantRange=False):
+def extract_slc(sensor, slantRange=False, removeNoise=False):
 #    sensor.configure()
     sensor.parse()
-    sensor.extractImage()
+    sensor.extractImage(removeNoise=removeNoise)
```
gshiroma, line 78: I think you cannot expect to have this ```removeNoise``` argument for every sensor (even though they are currently provided for all available GRD sensors). You may want to check it first by using ```inspect.getargspec()``` or just by calling it within a try/except.

gshiroma (commented):
This is indeed a great addition! Thank you, @vincentschut ! The code is well written and it handles the introduction of azimuth noise variation that was introduced in IPF version 2.9.0. From the first inspection, I just have one suggestion. I'll be reviewing this PR more in-depth in the next days.

vincentschut:
> Would you please attach an image of the computed noise for records (Perhaps before noise correction, after and the difference).

Sure.
Scene ID: S1B_IW_GRDH_1SDV_20200105T231133_20200105T231202_019688_02537D_8467, VV polarization
First image: without noise removal
S1B_IW_GRDH_1SDV_20200105T231133_20200105T231202_019688_02537D_8467_VV_no_noise_removed

Second image: with noise removal (the difference is hardly noticable, but it is there)
S1B_IW_GRDH_1SDV_20200105T231133_20200105T231202_019688_02537D_8467_VV_noise_removed

Third image: the noise that was removed (second image minus first image), range=0 to 0.01 (black to white)
S1B_IW_GRDH_1SDV_20200105T231133_20200105T231202_019688_02537D_8467_VV_noise

Note that the speckly noise you see in the third image in the water regions is due to setting negative values in the denoised image to 0, as is recommended in the ESA document I referenced above. You see them in the noise image, because I generated that by subtracting the denoised image from the non-denoised image.

components/isceobj/RtcProc/runPreprocessor.py
vincentschut: Right, will do.

vincentschut: Done in https://github.com/isce-framework/isce2/pull/192/commits/742d3b48c7bfbd615f5d8c70f8489a3e63b6fe79

avalentino:
@vincentschut thank you very much, it is a very interesting contribution.
I'm interested to difference between your implementation and the algorithm suggested in the ESA document. I have still not looked at the code closely but it would be nice if you could post an overview of main differences.

Regarding the preview images you posted I would suggest to adjust a little bit the colour scaling to have more dynamics on the sea area. You will have an image that is saturated and does not look nice on land but noise on the sea area will be more visible (it is also sufficient a small image portion).
Otherwise just choose a different image over sea.
After all the correction mainly makes sense on low backscattering regions.
I think that having a visual evidence of what this the improvement provided by the noise correction is very important.

vincentschut:
> I'm interested to difference between your implementation and the algorithm suggested in the ESA document. I have still not looked at the code closely but it would be nice if you could post an overview of main differences.

The only difference is an optimization, which doesn't affect the final result. In the ESA paper it is suggested that, for scenes that have azimuth noise vectors defined, to pre-calculate the noise matrix by looping over the azimuth vector blocks as defined in the noise xml file, and then for each block find the nearest range vector line(s) and use those. Whereas I am not looping over the blocks when creating the noise matrix, but I only loop over the blocks to add the outlines of that block to the final interpolators, allowing me to not having to calculate the entire (very big) noise matrix before applying it, but loop over lines and calculate only the noise matrix values for that line. This saves a lot of time and memory. Already in the ESA paper they mention that their example implementation suggestion is not ideal optimization-wise, and suggest people might implement it differently for efficiency reasons.

vincentschut:
> Regarding the preview images you posted I would suggest to adjust a little bit the colour scaling to have more dynamics on the sea area. You will have an image that is saturated and does not look nice on land but noise on the sea area will be more visible (it is also sufficient a small image portion).

I have changed the images in the comment above accordingly.

> After all the correction mainly makes sense on low backscattering regions.

Imho that is only partly true. In our experience, when creating mosaics form multiple scenes, without noise removal there'll be much stronger scene boundaries visible (where IW1 of one scene meets IW3 of another). (context: we (Satelligence) mainly use S1 to do change detection (deforestation mainly) or landcover classification on land. Thermal noise correction appeared to be a major improvement to some of our results, e.g. oil palm plantation classification in West-Africa. We used SNAP then, but are now considering moving to ISCE instead of SNAP).

> I think that having a visual evidence of what this the improvement provided by the noise correction is very important.

Totally agree. Please let me know if you need more (images or figures). (btw starting tomorrow I'll be mostly away for 1 week, I'll be back at work starting Oct. 26).

avalentino:
@vincentschut thanks for your clarification about the algorithm implementation.
Regarding the image, unfortunately it seems that the ambiguous signal coming form land does not allow to observe the noise pattern clearly. Probably an image acquired mainly on sea would serve better to the purpose.

components/isceobj/RtcProc/runPreprocessor.py
gshiroma: Thank you, @vincentschut !

gshiroma (approved):
Thank you for addressing our comments , @vincentschut . The figures were very helpful. As I mentioned before, the code is clear and well-written. I couldn't find any issues. At first, I thought it was strange to repeat the azimuth LUT values at ```block_range_start``` and ```block_range_end``` when defining the az. interpolation in the range direction. It seemed more logical to me to use the center (range position) of the block. However, I noticed that the S-1 documentation recommends something similar (i.e. to repeat the azimuth LUT values between ```block_range_start``` and ```block_range_end```), so, I guess it's fine. The range LUT computation seems fine as well although the ```RectBivariateSpline``` was substituted by a ```InterpolatedUnivariateSpline``` followed by an ```interp1d```, but I don't see problems with that. A very small suggestion would be to avoid using the term range in the sense of bounds/extent as it could be confused with range/azimuth. But in this case, the developers would understand the difference. In my opinion, the code is ready for merging. This is definitely a great addition! Thank you again!

PR #193, opened 2020-10-07
topsStack: add --num-proc for step topo

yunjunz:
This PR includes the following changes:

+ add `stackSentinel.py --num-proc` to pass the number of parallel processes to topo.py, so that mp.Pool() can be independent of OMP_NUM_THREADS, following the discussion at https://github.com/isce-framework/isce2/commit/f06d22e02c43e076e66b19ddcc76506bbefcae2a

+ Stack: check existing orbit with glob instead of os.path.exist(dir)

+ stackSentinel: fix the missing pair between the reference date and the new acquisitions, if the reference date is the most recent "num_conn" acquisition.

Please see the code change for each commit separately, since there is one commit that aligns linebreaks in stackSentinel.createParser().

### DIFF
--- a/contrib/stack/topsStack/Stack.py
+++ b/contrib/stack/topsStack/Stack.py
@@ -75,6 +75,7 @@ class config(object):
         self.f.write('reference : ' + self.outDir + '\n')
         self.f.write('dem : ' + self.dem + '\n')
         self.f.write('geom_referenceDir : ' + self.geom_referenceDir + '\n')
+        self.f.write('numProcess : ' + str(self.numProcess) + '\n')
         self.f.write('##########################' + '\n')
 
     def geo2rdr(self,function):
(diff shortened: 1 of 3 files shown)


hfattahi (approved):

cmarshak:
Can I use this functionality in topsApp?

dbekaert:
@yunjunz @hfattahi understand this is to overwrite the OMP_NUM_THREADS. Aware of the topsStack/TopsApp differences. Is there a configurable parameter that can be controlled to speed up topo processing for CPU for topsApp or is this only under OMP_NUM_THREADS?

yunjunz:
I don't think the current topsApp has the option to use multiple processes, the only control parameter to speed up is the OMP_NUM_THREADS for multiple threads. 

It should not be difficult to modify the code below to add the multiple processes though:
https://github.com/isce-framework/isce2/blob/def109815df5f724c5dc653878314577e3505c48/components/isceobj/TopsProc/runTopo.py#L25

dbekaert:
@ehavazli Could you take a stab at this under ACCESS? Specifically this would be migrating #99 into TopsApp, specifically the function as identified above by @yunjunz

yunjunz:
@dbekaert I am using topsStack all the time, and do not really have much experience on topsApp. Based on the runtime noted in #99, yes I think it would be useful to have it in topsApp as well.

PR #195, opened 2020-10-22
topsStack: ignore coreg_secondary images with inconsistent num of bursts

yunjunz:
This PR adds checking in `extractCommonValidRegion.py` and `mergeBursts.py` to skip images in the `coreg_secondarys` folder that have different numbers of bursts than the reference. Thanks to @piyushrpt for guidance on debugging and fixing.

The PR also rename the `merge` step run file in slc workflow to be consistent with the one from interferogram/correlation workflow.


hfattahi (approved):
LGTM

PR #202, opened 2020-11-03
fix S1 GRD thermal noise removal for AzimuthNoiseVectors with a lineCount of 1.

vincentschut:
Fix for the recently implemented thermal noise correction for S1 GRD data, for cases where an AzimuthNoiseVector has only 1 value. The current implementation tries to fit a InterpolatedUnivariateSpline to that single value, and obviously fails. This fix just uses the single value in those cases, without an intermediate InterpolatedUnivariateSpline.

Example scene ID with such a 'feature': S1B_IW_GRDH_1SDV_20200129T231132_20200129T231201_020038_025EA8_46CD


gshiroma (approved):
LGTM. Thank you, @vincentschut , for addressing this case.

PR #208, opened 2020-11-12
PyCuAmpcor updates:

lijun99:
* added a README.md for installation/user guide/procedures

  * modified the cuDenseOffsets.py
     * expose more options from the CUDA/C++ program
     * add an option for varying gross offset input
     * clarify the parameter definitions

  * removed old SlcImage implementation and cublas dependence

  * modified cuSincOversampler
    * to be consistent with cpu version
    * fix an issue when the extraction of the search window is not around the center

  * added a debug mode to output intermediate results

  * enable cuda error checking for both Debug/Release build types

  * corrected the code to extract raw images when the correlation surface peak is close to edges

  * Move utf-8 decoding step inside cython extension

    The cython setters take python strings, but the getters return
    python bytes, so this makes the types match up. I went with regular
    python strings for the interface since they are more common at the
    python level, so the encoding/decoding is now an implementation detail
    of the cython extension.

Contributed by lijun99, rtburns-jpl, vbrancat, mzzhong


contrib/PyCuAmpcor/README.md
```diff
@@ -0,0 +1,442 @@
...
+
+## 6. Additional Notes
+
+### 6.1 Sinc Oversampler
+
+The since oversampler/interpolator may be selected to oversample the correlation surface with *--corr-osm=1* in *cuDenseOffsets.py*, or *objOffset.corrSurfaceOverSamplingMethod=1*.
+
+The sinc interpolating formula is defined as
+
+ $$x(t) = \sum_{n=-\infty}^{\infty} x_n f( \Omega_c t-n )$$
```
rtburns-jpl, line 424: It looks like these tex equations aren't being rendered properly in the browser. What do you think about something like this? https://gist.github.com/a-rodin/fef3f543412d6e1ec5b6cf55bf197d7b

contrib/PyCuAmpcor/src/cuAmpcorChunk.cu
```diff
@@ -17,122 +17,221 @@ void cuAmpcorChunk::run(int idxDown_, int idxAcross_)
...
         stream);
-    //offsetInit->debuginfo(stream);
-    //offsetZoomIn->debuginfo(stream);
-    //offsetFinal->debuginfo(stream);
+
+// #ifdef CUAMPCOR_DEBUG
+    // std::cout << "Offsets: Oversampled and Final)\n";
+    // offsetZoomIn->debuginfo(stream);
+    // offsetFinal->debuginfo(stream);
+// #endif
```
rtburns-jpl, line 229: I think this block can be uncommented since it's already ifdef'd out

contrib/PyCuAmpcor/src/cuAmpcorChunk.cu
```diff
@@ -17,122 +17,221 @@ void cuAmpcorChunk::run(int idxDown_, int idxAcross_)
...
         param->halfZoomWindowSizeRaw,  // new range
         param->halfZoomWindowSizeRaw,
         stream);
-    //offsetInit->debuginfo(stream);
+
+#ifdef CUAMPCOR_DEBUG
+    // std::cout << "max location adjusted if close to boundary\n";
+    // offsetInit->debuginfo(stream);
+    // std::cout << "and the shift of the center\n";
+    // maxLocShift->debuginfo(stream);
```
rtburns-jpl, line 120: Maybe uncomment these, since they're already ifdef'd? Seems like they would be useful when debugging

contrib/PyCuAmpcor/src/SConscript
```diff
@@ -16,6 +17,10 @@ listFiles = ['GDALImage.cu', 'cuArrays.cu', 'cuArraysCopy.cu',
 
 lib = envPyCuAmpcor.SharedLibrary(target = 'PyCuAmpcor', source= listFiles, SHLIBPREFIX='')
 
+# add gdal include path
+gdal_cflags = subprocess.check_output('gdal-config --cflags', shell=True)[:-1].decode('utf-8')
```
rtburns-jpl, line 21: How does this behave when gdal-config does not exist? I hope it's able to proceed in that case

rtburns-jpl (changes requested):
Awesome! I'll try to do a more in-depth review later and run the code but what I've seen so far looks great. Have a couple minor suggestions.

The README looks really great too - permalink for previewing in browser: https://github.com/lijun99/isce2/blob/a393282/contrib/PyCuAmpcor/README.md

I see you opted for a CUAMPCOR_DEBUG flag in addition to NDEBUG. That's a good idea, since there's certainly a use case for testing in Debug mode but without the very expensive image dumps. Might be good to advertise this feature in the README.

I'm very anti-commented-out-code, so I would say to remove any of those.

Looking forward to testing this out :)

contrib/PyCuAmpcor/README.md
lijun99: That seems to be a common problem with markdown on github. I don't really like the idea of using an external generator, unless you strongly favor it. Or, I can use an ipynb which renders tex formula correctly on github.

contrib/PyCuAmpcor/src/cuAmpcorChunk.cu
lijun99: I will clean that.

contrib/PyCuAmpcor/src/cuAmpcorChunk.cu
lijun99: I will clean that.

contrib/PyCuAmpcor/src/SConscript
lijun99: gdal-config is used by scons to check the existence of gdal. So it should be OK. Otherwise, it will be good to define an env['GDAL_PATH'] so that I can pull from.

lijun99:
Thanks for the comments. I will clean the commented code. For the debug, in debug.h, I use NDEBUG to turn on/off CUAMPCOR_DEBUG. For CUDA_ERROR_CHECK, it's better to keep it on all the time, otherwise, the program won't be able to detect even memory allocation errors.

contrib/PyCuAmpcor/README.md
rtburns-jpl: Yeah I agree it's not ideal, just thought I'd bring it up since it might be helpful to those reading it online. Ipynb could be a nice option if it's not too much trouble. I'll defer to your judgement on this.

lijun99:
Please check the new commit for cleaned code. I also went through the entire code, adding docstrings/notes and deleting some obsolete code.  Hopefully, the code is now more readable and easier to understand.

rtburns-jpl (approved):
Awesome, code cleanup looks great and is very readable. @vbrancat has been showing great results with this

lijun99:
The new commit (6d9f088) includes changes to cuDenseOffsets.py to work properly with input varying gross offsets, contributed by @mzzhong.

rtburns-jpl:
Had a chance to test this out today and it worked great! Planning to merge later this week.

PR #209, opened 2020-11-30
Clean up and update grossOffsets

mzzhong:
1. There exists two versions of cuDenseOffsets.py (the python driver of PyCuAmpcor) in ISCE2. One is at "contrib/PyCuAmpcor/examples/cuDenseOffsets.py" and the other one is at "contrib/stack/topsStack/cuDenseOffsets.py". Only one version should be kept. After some discussion, we want to keep all PyCuAmpcor specific code including "cuDenseOffsets.py" within "contrib/PyCuAmpcor/examples", so we remove the other version in the TOPS stack processor.

2. grossOffsets.py is to estimate from the Antarctica ice velocity model the predicted pixel offsets given the viewing geometry. It is supposed to work with PyCuAmpcor by providing the gross offsets. The code is at "contrib/stack/topsStack/grossOffsets.py". For the same reason as above, we move the code to "contrib/PyCuAmpcor/examples".

3. The existing grossOffsets.py is outdated. Here, I provide a new version of grossOffsets.py for public use. It is independent of PyCuAmpcor and writes the results to raw binary file. The raw binary file will be read by the latest cuDenseOffsets.py (to be updated by Lijun's PR: https://github.com/isce-framework/isce2/pull/208) and feed to PyCuAmpcor. I have tested grossOffsets.py.


mzzhong:
I am closing this PR and re-submit it to Lijun's ISCE2 repo. We will make a new PR from there including updates of PyCuAmpcor as well as this change here.

PR #215, opened 2020-12-31
stackSentinel.py:using geopandas to calculate intersection between bbox and SAFE file coverage region

ranneylxr:
using geopandas and shapely package to calculate whether there are intersections between bbox and sentinel SAFE file.

### DIFF
--- a/contrib/stack/topsStack/stackSentinel.py
+++ b/contrib/stack/topsStack/stackSentinel.py
@@ -11,6 +11,9 @@ import datetime
 import time
 import numpy as np
 
+import geopandas as gpd
+from shapely.geometry import Point, Polygon
+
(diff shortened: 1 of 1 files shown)


contrib/stack/topsStack/stackSentinel.py
```diff
@@ -11,6 +11,9 @@
 import time
 import numpy as np
 
+import geopandas as gpd
+from shapely.geometry import Point, Polygon
```
rtburns-jpl, line 15: Can these imports be moved inside the `generate_geopolygon` function? These python packages aren't formal prerequisites of isce2 so I'd like to avoid ImportErrors when they aren't otherwise needed

PR #216, opened 2020-12-31
using geopandas and shapely package to calculate whether there are intersections between bbox and sentinel SAFE file.

ranneylxr:
Using geopandas and shapely package to calculate whether there are intersections between bbox and sentinel SAFE file.
And remove "import geopandas and shapely" part into generate_geopolygon function.

### DIFF
--- a/contrib/stack/topsStack/stackSentinel.py
+++ b/contrib/stack/topsStack/stackSentinel.py
@@ -11,6 +11,9 @@ import datetime
 import time
 import numpy as np
 
+import geopandas as gpd
+from shapely.geometry import Point, Polygon
+
 # suppress matplotlib DEBUG message
 from matplotlib.path import Path as Path
 import logging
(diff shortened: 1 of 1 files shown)


piyushrpt:
Just noting that introduces a new dependency on geopandas. The same can be done with `polygon1.intersects(polygon2)` in shapely?

yunjunz (changes requested):
Thank you @ranneylxr for improving this. I would second @piyushrpt's suggestion on removing `geopandas` as it seems like just using shapely should be enough. [Here](https://gis.stackexchange.com/questions/90055/finding-if-two-polygons-intersect-in-python) is another similar topic on StackExchange. Otherwise, adding a new dependency would result in quite some changes on the installation side.

ranneylxr:
Hi @piyushrpt and @yunjunz . Thank you for your valuable suggestions. Now the script only uses shapely package. Please check.

contrib/stack/topsStack/stackSentinel.py
```diff
@@ -196,6 +196,27 @@ def cmdLineParse(iargs = None):
...
+    p4 = Point(bbox[3][0], bbox[3][1])
+
+    np1 = (p1.coords.xy[0][0], p1.coords.xy[1][0])
+    np2 = (p2.coords.xy[0][0], p2.coords.xy[1][0])
+    np3 = (p3.coords.xy[0][0], p3.coords.xy[1][0])
+    np4 = (p4.coords.xy[0][0], p4.coords.xy[1][0])
+
+    bb_polygon = Polygon([np1, np2, np3, np4])
+
+    return bb_polygon
```
rtburns-jpl, line 218: I think using a list comprehension makes this a bit simpler
```suggestion
    points = [Point(bbox[i][0], bbox[i][1]) for i in range(4)]
    return Polygon([(p.coords.xy[0][0], p.coords.xy[1][0]) for p in points])
```

yunjunz:
Since `matplotlib.path.Path` is not used anymore, could you please also remove [stackSentinel.py#L14-L18](https://github.com/isce-framework/isce2/blob/main/contrib/stack/topsStack/stackSentinel.py#L14-L18)?

yunjunz (approved):
Looks good and clean. Thank you @ranneylxr!

hfattahi (approved):
LGTM. Thank you.

PR #219, opened 2021-01-06
Modernize gdal imports

rtburns-jpl:
GDAL has dropped support for the older toplevel `import gdal`
in version 3.2, now requiring `from osgeo import gdal`.


hfattahi (approved):

lijun99:
There is another one: 
components/isceobj/StripmapProc/runDispersive.py:import os,gdal

rtburns-jpl:
Thanks, fixed in https://github.com/isce-framework/isce2/commit/0d5b94efadbc1dcfcb7fea91e6152db7cfdb3e8d

Issue #182, opened 2020-08-24
TypeError: runGeocode() takes 1 positional argument but 3 were given


HBaldwin3:
Not sure how to resolve this error (posted to user forum as well: http://earthdef.caltech.edu/boards/4/topics/3295?r=3553#message-3553)

https://github.com/isce-framework/isce2/blob/b13d06aa3b981119d32c9da7dd1bb244622bed1e/components/isceobj/RtcProc/Factories.py#L40

![image](https://user-images.githubusercontent.com/49163135/91048065-b7f5dd80-e5e0-11ea-9262-53574ff2339a.png)

bjmarfito:
This could be possibly a problem on the rtcApp.py where on line nos. 427 to 428 shows a function needing three arguments instead of just one argument. 
[https://github.com/isce-framework/isce2/blob/f65a51c80c2605001ebe6f97dd98f7b45bc1d919/applications/rtcApp.py#L427](url)
[https://github.com/isce-framework/isce2/blob/f65a51c80c2605001ebe6f97dd98f7b45bc1d919/applications/rtcApp.py#L428](url)

This could possibly be solved by removing the args=(self.geocode_list, self.geocode_bbox) on rtcApp.py or add those arguments on the runGeocode.py on the RtcProc folder.

bjmarfito:
I checked the runGeocode.py on the RtcProc folder. There is no used if the code will incorporate self.geocode_list, self.geocode_bbox. I suggest removing it on the rtcApp.py.

bjmarfito:
I generated a PR to resolve this issue. [https://github.com/isce-framework/isce2/pull/220](url)

PR #220, opened 2021-01-09
Update rtcApp.py

bjmarfito:
I recommend to remove the arguments args=(self.geocode_list, self.geocode_bbox) since it passes the arguments on the runGeocode.py of the RtcProc which prevents geocoding of the sigma nought product generated from the rtcApp.py. This could be added later if someone would incorporate the geocode_list and geocode_bbox on the runGeocode.py later.

### DIFF
--- a/applications/rtcApp.py
+++ b/applications/rtcApp.py
@@ -424,8 +424,7 @@ class GRDSAR(Application):
         self.step('normalize', func=self.runNormalize)
 
         # Geocode
-        self.step('geocode', func=self.runGeocode,
-                args=(self.geocode_list, self.geocode_bbox))
+        self.step('geocode', func=self.runGeocode)
 
         return None
 


rtburns-jpl:
This looks correct as far as I can tell. I see some other objects have a runGeocode method which takes additional arguments, but the one in RtcProc used by rtcApp.py does not. Maybe @HBaldwin3 can verify that this fixes issue https://github.com/isce-framework/isce2/issues/182?

scottyhq:
I can confirm this fixes the linked issue. [Log here](https://github.com/relativeorbit/isce2-rtc/runs/4051024585?check_suite_focus=true) should be public for a while... Would be great to merge as rtcApp currently errors when hitting the geocode step. Might be worth also removing references to geocode_list and geocode_bbox from rtcApp.py to avoid user confusion:

https://github.com/isce-framework/isce2/blob/816459c9541f601d607b45dcf6011c60d48c1a88/applications/rtcApp.py#L186 

and https://github.com/isce-framework/isce2/blob/816459c9541f601d607b45dcf6011c60d48c1a88/applications/rtcApp.py#L94 

and adding that functionality in the future...

rtburns-jpl:
Great, thanks for confirming. Let's go ahead and merge this then - your other suggestions sound good for future steps.

PR #221, opened 2021-01-13
add some includes for c header files

lijun99:
Apple clang 12.0.0 reports errors for implicit declarations. This PR fixes some missing header file includes. Now ISCE2 can be compiled with the newest Bug Sur + Xcode 12.3, together with brew installed gfortran and conda. 

PS: my editor automatically removes the extra spaces at the end of line. For io.c, the only change is adding #include <string.h> for calling strlen().


rtburns-jpl:
Thanks! Once I have some time to look into reworking our CI, I'd like to add a clang build job to catch things like this that don't show up with gcc.

PR #224, opened 2021-01-28
Updates to PyCuAmpcor

lijun99:
This PR includes various updates to PyCuAmpcor:

1. Use the template size to scale the covariance estimation of the correlation surface (as a measure of errors for the offset fields). It is now consistent with RIOPAC. Contributed by @mzzhong.

2. Add/Modify the grossOffsets.py, which provides an estimate of (gross) offset fields with Antarctica velocity model. The generated gross offset fields can be used by ampcor as varying gross offsets to reduce the search range. This change was reported early as PR #209. Contributed by @mzzhong. 

3. Bug fix: fix the helper string in cuDenseOffsets.py


rtburns-jpl (approved):
Thanks for rolling these up into one PR, I think this is actually easier because I wasn't sure which order I should be looking at the changes. LGTM

PR #227, opened 2021-02-02
Add cmake files for GPUresampslc & GPUtopozero

rtburns-jpl:
Thanks to @lijun99 for pointing out that these were missing!
@yuankailiu, please let me know if this fixes your issue.


yuankailiu:
Thanks for the updates!

That works for me now. 

But for the below CMake List files under [isce2/components/components/zerodop](https://github.com/isce-framework/isce2/tree/main/components/zerodop): 
```
GPUtopozero/CMakeLists.txt
GPUgeo2rdr/CMakeLists.txt
```

we need to add this line at the very beginning:
``` js
add_definitions(-DGPU_ACC_ENABLED)
```

PR #231, opened 2021-02-09
GPU modules: add GPU_ACC_ENABLED flag to cmake

lijun99:
GPU modules under components/zerodop seem to include both CPU and GPU versions and use a GPU_ACC_ENABLED flag to control which version to run. Therefore, we need to add this flag to cmake, as also included in SCons. 

GPUtopozero, GPUgeo2rdr, as included in Alos2Proc, TopsProc, have been tested to run as expected. 

GPUresampslc is not included in any script, and has not been tested. 

GPUampcor is out-dated and has not been tested.


rtburns-jpl (approved):
Thanks! You are correct, `target_compile_definitions` is the recommended way to do this.

PR #232, opened 2021-02-11
Stack minor modification

### LINKED ISSUE #230: problem with the commit  #225: 'Bash script edits to make stackSentinel.py faster in places'
This change has proposed a few difficulties in job submission on HPC.
It was agreed to be optional but the way it is implemented is based on existing `num_process` argumet of stackSentinel.py
This argument was supposed to be used in `multiprocessing` module to speedup topo.py
The problem is that we would like it to be 1 (not applied) for ifgrams and unwrapping steps, as it is proposed but we definitely need it to be a higher value for topo.py
run files should not deal with parallelization as there are so many ways to do that, but if you prefer to keep this change, we should consider to use another argument, maybe a new one other than `num_process`
@yunjunz , @hfattahi  what is your opinion?

mirzaees:
Proposed changes:

- Added a new argument to stackSentinel.py named `num_proc4topo` for topo.py  to solve the conflicting issue of parallel processing 
- Change the run files numbering style in stackStripMap.py to be consistent with stackSentinel.py

### DIFF
--- a/contrib/stack/topsStack/Stack.py
+++ b/contrib/stack/topsStack/Stack.py
@@ -333,7 +333,6 @@ class run(object):
         configObj.outDir = os.path.join(self.work_dir, 'reference')
         configObj.geom_referenceDir = os.path.join(self.work_dir, 'geom_reference')
         configObj.dem = os.path.join(self.work_dir, configObj.dem)
-        configObj.numProcess = self.numProcess4topo
         configObj.Sentinel1_TOPS('[Function-1]')
         configObj.topo('[Function-2]')
         configObj.finalize()
--- a/contrib/stack/topsStack/Stack.py
+++ b/contrib/stack/topsStack/Stack.py
@@ -76,7 +76,7 @@ class config(object):
         self.f.write('reference : ' + self.outDir + '\n')
         self.f.write('dem : ' + self.dem + '\n')
         self.f.write('geom_referenceDir : ' + self.geom_referenceDir + '\n')
-        self.f.write('numProcess : ' + str(self.numProcess) + '\n')
+        self.f.write('numProcess : ' + str(self.numProcess4topo) + '\n')
         self.f.write('##########################' + '\n')
 
     def geo2rdr(self,function):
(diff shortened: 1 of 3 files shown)


contrib/stack/topsStack/Stack.py
```diff
@@ -333,7 +333,7 @@ def unpackStackReferenceSLC(self, safe_dict):
         configObj.outDir = os.path.join(self.work_dir, 'reference')
         configObj.geom_referenceDir = os.path.join(self.work_dir, 'geom_reference')
         configObj.dem = os.path.join(self.work_dir, configObj.dem)
-        configObj.numProcess = self.numProcess
+        configObj.numProcess = self.numProcess4topo
```
yunjunz, line 336: Since `self/configObj.numProcess4topo` is used in `topo()`, we should not need this line anymore.

contrib/stack/topsStack/stackSentinel.py
```diff
@@ -166,8 +166,11 @@ def createParser():
     parser.add_argument('-useGPU', '--useGPU', dest='useGPU',action='store_true', default=False,
                         help='Allow App to use GPU when available')
 
-    parser.add_argument('--num-proc', '--num-process', dest='numProcess', type=int, default=1,
-                        help='number of parallel processes (where applicable) (default: %(default)s).')
+    parser.add_argument('--num_proc', '--num_process', dest='numProcess', type=int, default=1,
+                        help='number of tasks running in parallel in each run file (default: %(default)s).')
+
+    parser.add_argument('--num_proc4topo', '--num_process4topo', dest='numProcess4topo', type=int, default=1,
+                        help='number of parallel processes (for topo only) (default: %(default)s).')
```
yunjunz, line 173: Not a big deal at all, but it would be more convenient for non-HPC users if we could tie `numProcess4topo` to `numProcess` in `cmdLineParse()` if the later is specified in command line while the former is not, such as below:

```python
    if any(i in iargs for i in ['--num_proc', '--num_process']) and all(i not in iargs for i in ['--num_proc4topo', '--num_process4topo']):
        inps.numProcess4topo = inps.numProcess
```

yunjunz:
#230

contrib/stack/topsStack/stackSentinel.py
mirzaees: @yunjunz  please check if it is working fine as you suggested

yunjunz (approved):
LGTM, thank you @mirzaees.

PR #233, opened 2021-02-11
Update CMakeLists.txt

bjmarfito:
I made some changes to the CMakeLists.txt since I can't process ALOS-2 data due to the absence of alos2_slc on the db folder. Later, I've also found out that all of the folders inside the db, with the exception to the alos folder, are not included during installation.


rtburns-jpl:
This seems incomplete - don't you need CMakeLists.txt files in each of those subdirectories?

```
...

CMake Error at components/isceobj/Sensor/db/CMakeLists.txt:2 (add_subdirectory):
  The source directory

    /root/project/src/components/isceobj/Sensor/db/alos2_slc

  does not contain a CMakeLists.txt file.


CMake Error at components/isceobj/Sensor/db/CMakeLists.txt:3 (add_subdirectory):
  The source directory

    /root/project/src/components/isceobj/Sensor/db/alos_slc

  does not contain a CMakeLists.txt file.


CMake Error at components/isceobj/Sensor/db/CMakeLists.txt:4 (add_subdirectory):
  The source directory

    /root/project/src/components/isceobj/Sensor/db/ers

  does not contain a CMakeLists.txt file.

...
```

bjmarfito:
Thank you for spotting my mistake. I'll upload the CMakeLists.txt file for each of the subdirectories later.

PR #235, opened 2021-02-12
Fix Cmake lists for db folder

bjmarfito:
I've made some changes based on my previous request (#233) and inputs by @rtburns-jpl.


rtburns-jpl (approved):
Thanks! Fyi, in the future you can push to your branch and the PR will be updated accordingly - no need to create a new PR for each patchset.

bjmarfito:
This is noted.

> Thanks! Fyi, in the future you can push to your branch and the PR will be updated accordingly - no need to create a new PR for each patchset.

PR #237, opened 2021-02-20
Fix missing folders and files during CMake installation of GRD files

bjmarfito:
I've added some of the missing codes and a CMakeLists.txt for processing of GRD files. I've tested the rtcApp.py and errors show up due to the necessary folder and files not included during installation.


rtburns-jpl:
Thanks!

PR #241, opened 2021-02-28
Resolved the issue on the rdrdem_offset stage

### LINKED ISSUE #239: Processing stopped during the rdrdem_offset stage
I've processed ALOS-2 stripmap images to generate DInSAR images using the alos2App.py. Then, during the rdrdem_offset stage an error occurred halting the data processing. I've attached the screenshot of the error message for reference. I've used the latest version of ISCE2 and compiled it using CMake for this. I've also processed the images using the Conda version of ISCE2 and it processed flawlessly.
error

bjmarfito:
After examining the error and the runRdrDemOffset.py, I've realized that the cause of the problem on https://github.com/isce-framework/isce2/issues/239 was the fitoff.f which causes the segmentation error. After some experimentation, I've figured out that the OpenMP_Fortran is causing the issue. I removed it from the CMakeLists.txt and recompiled the code. The rdrdem_offset stage ran successfully.

### DIFF
--- a/contrib/alos2proc_f/CMakeLists.txt
+++ b/contrib/alos2proc_f/CMakeLists.txt
@@ -27,7 +27,8 @@ cython_add_module(alos2proc_f
     )
 
 target_include_directories(alos2proc_f PUBLIC include)
-target_link_libraries(alos2proc_f PUBLIC FFTW::Float OpenMP::OpenMP_Fortran)
+target_link_libraries(alos2proc_f PUBLIC 
+    FFTW::Float)
 
 InstallSameDir(
     alos2proc_f


rtburns-jpl:
@CunrenLiang, does this make sense to you? I'm not sure why OpenMP would be causing issues here.

bjmarfito:
To give you some context of the issue which could probably help in solving the problem, the top screenshot above shows line no. 238 (highlighted code) of the runRdrDemOffset.py which causes the processing error. Once, I've run it manually using the terminal it shows a segmentation error as shown in the middle screenshot which is caused when the code executes the fitoff.f. The error is caused by the fitoff.f when executed. I'm also attaching a screenshot (bottom screenshot) when the fitoff.f is running properly.

error_4
Screen Shot 2021-03-01 at 7 12 32 AM
Screen Shot 2021-03-01 at 7 13 58 AM

CunrenLiang:
Thanks for posting the details. For the 3rd (1.5) and 4th (0.5) parameters of fit off, can you try other values to see if the program can run successfully?
e.g.
3rd parameter: 0.4,  0.5,  0.6
4th parameter: 1.5,  1.4,  1.3,  1.2,  1.1,  1.0,  0.9

CunrenLiang:
Or can you also share ampcor.off?

bjmarfito:
The same error occurs even if I changed the 3rd and 4th parameters. Pls. see the screenshot below.
Screen Shot 2021-03-01 at 10 41 25 AM



> Thanks for posting the details. For the 3rd (1.5) and 4th (0.5) parameters of fit off, can you try other values to see if the program can run successfully?
> e.g.
> 3rd parameter: 0.4, 0.5, 0.6
> 4th parameter: 1.5, 1.4, 1.3, 1.2, 1.1, 1.0, 0.9

bjmarfito:
> Or can you also share ampcor.off?

Here is the ampcor.off.

[ampcor.off.zip](https://github.com/isce-framework/isce2/files/6058828/ampcor.off.zip)

CunrenLiang:
@bjmarfito may be right @rtburns-jpl. I tried two versions of the compiled fitoff program. One compiled with scons in ISCE, and the other one is a standalone program compiled by gfortran.  Both do not have the reported issue.
the contrib/alos2proc_f module does not requrie OpenMP_Fortran anyway, so it can be removed from cmake script.

rtburns-jpl:
Ok, great! Thanks to you both for working out this issue.

Issue #243, opened 2021-03-01
Ampcor to support URL as file input


whyjz:
Hi, 

I have been using ISCE, particularly the Ampcor component, for my feature tracking package [CARST](https://github.com/whyjz/CARST). I run feature tracking using both optical and SAR images. 

My question is: when the input file is stored on a remote server and is accessed via a URL, e.g., a GeoTIFF file on the Amazon web service, the `iscesys.ImageApi.DataAccessor` cannot open it because it needs a VRT file that is stored on the same path as the input file, and we usually don't have the write permission on a remote server (i.e. not able to create the necessary VRT file on the same path). 

One workaround is to make it more flexible for generating/selecting the VRT file path (for example, by changing line 145 of `components/iscesys/ImageApi/DataAccessor/DataAccessorPy.py`) I just made some modification for that and made sure it's working. I wonder if that's okay for me to make a PR, or if you have a better idea to deal with this issue. Thanks!

piyushrpt:
You should be able to create a local VRT file with source pointing to remote resource and corresponding xml file, and ISCE should be able to work with it - for reading data directly. Yes, please issue a PR - would be good to know if any other changes were needed.

PR #245, opened 2021-03-03
Add URL support to DataAccessor

whyjz:
As discussed in #243, I edited the rule for finding the VRT -- if the file name is actually a URL, the Data Accessor will read the corresponding VRT file (that points to the remote, original file) on the local machine. It should be compatible with almost all the ICSE use cases. Let me know if you have any questions about this change.

Thanks!

### DIFF
--- a/components/iscesys/ImageApi/DataAccessor/DataAccessorPy.py
+++ b/components/iscesys/ImageApi/DataAccessor/DataAccessorPy.py
@@ -142,7 +142,12 @@ class DataAccessor(object):
         caster = '' or self.caster
         filename = self.filename
         scheme = self.scheme
-        self.extraFilename = self.filename + '.' + self._extra_reader
+        #if the filename is a URL, the extraFilename should indicate the file from the local machine
+        #instead of from the remote server.
+        if self.filename.startswith('http'):
+            self.extraFilename = os.path.basename(self.filename) + '.' + self._extra_reader
+        else:
+            self.extraFilename = self.filename + '.' + self._extra_reader
 
         if self._accessor is None:#to avoid creating duplicates
             selection = self.methodSelector()
--- a/components/iscesys/ImageApi/DataAccessor/DataAccessorPy.py
+++ b/components/iscesys/ImageApi/DataAccessor/DataAccessorPy.py
@@ -144,7 +144,7 @@ class DataAccessor(object):
         scheme = self.scheme
         #if the filename is a URL, the extraFilename should indicate the file from the local machine
         #instead of from the remote server.
-        if self.filename.startswith('http'):
+        if self.filename.startswith('http://') or self.filename.startswith('https://'):
             self.extraFilename = os.path.basename(self.filename) + '.' + self._extra_reader
         else:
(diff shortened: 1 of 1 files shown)


components/iscesys/ImageApi/DataAccessor/DataAccessorPy.py
```diff
@@ -142,7 +142,12 @@ def createAccessor(self):
         caster = '' or self.caster
         filename = self.filename
         scheme = self.scheme
-        self.extraFilename = self.filename + '.' + self._extra_reader
+        #if the filename is a URL, the extraFilename should indicate the file from the local machine
+        #instead of from the remote server.
+        if self.filename.startswith('http'):
```
rtburns-jpl, line 147:
```suggestion
        if self.filename.startswith('http://') or self.filename.startswith('https://'):
```
It may be possible someone's filename starts with `http`, but I wouldn't expect a local filename to contain `//`

rtburns-jpl (commented):
@piyushrpt do you think this is a reasonable assumption to make? Can you imagine a situation where someone would want the extraFilename to be remote as well?

components/iscesys/ImageApi/DataAccessor/DataAccessorPy.py
piyushrpt: Why not create a .vrt and .xml and have source in vrt to point to remote? That way you have a record of what resource was actually used? That way you are not restricted to something thats just http / https. You can throw in anything vsis3 / vsiaz/ vsicurl etc. This approach still seems to need a .vrt. Just using  basename of .vrt should be enough when it points to other things?

whyjz: @piyushrpt I did create a .vrt that points to a remote geotiff. In my use case, the XML is not needed. Here's an example of what would happen if running this snippet using the original code version:

```
import isce
from isceobj.Image.Image import Image
obj = Image()
obj.setFilename("https://download.osgeo.org/geotiff/samples/GeogToWGS84GeoKey/GeogToWGS84GeoKey5.tif")
obj.setWidth(101)
obj.setDataType('SHORT')
obj.setCaster('read', 'FLOAT')
obj.createImage()
# ... (continue to some Ampcor code)
```
```
GDAL open (R): https://download.osgeo.org/geotiff/samples/GeogToWGS84GeoKey/GeogToWGS84GeoKey5.tif.vrt
ERROR 1: HTTP error code : 404
Error. Cannot open the file https://download.osgeo.org/geotiff/samples/GeogToWGS84GeoKey/GeogToWGS84GeoKey5.tif.vrt in read mode.
Error in file /home/whyj/Software/isce2-2.4.2-build/isce/components/iscesys/ImageApi/InterleavedAccessor/src/GDALAccessor.cpp at line 77 Exiting
```
You can see that the ISCE's default behavior is to read a VRT that is at the same location as the original file, even you have created a VRT on your local machine. If we replace the geotiff file name with the local vrt, ISCE will try to gdal open `.vrt.vrt`, which does not exist.

Issue #249, opened 2021-03-05
stackSentinel.py couldn't recognize the orbit EOF file and download it


mohseniaref-InSAR:
Dear all,

It seems stackSentinel.py can not recognize downloaded EOF in folder and try to download file . I really appreciate if you can help me in order to solve this issue
```
stackSentinel.py  -s /raid/maref/InSAR/S1/desc_tr10_f -o /raid-manaslu/maref/InSAR/ORBIT/aux_poeorb/  -a /raid/maref/InSAR/ORBIT/aux_ins/ -d demLat_S28_S22_Lon_W070_W062.dem.wgs84 -m 20150814  -c all -W slc  -b '-25.3 -24.0 -66.2 -65.2' 2>&1 | tee stackSentinel.log
This is the Open Source version of ISCE.
Some of the workflows depend on a separate licensed package.
To obtain the licensed package, please make a request for ISCE
through the website: https://download.jpl.nasa.gov/ops/request/index.cfm.
Alternatively, if you are a member, or can become a member of WinSAR
you may be able to obtain access to a version of the licensed sofware at
https://winsar.unavco.org/software/isce
Number of SAFE files found: 280
2021-03-05 23:33:35,219 - shapely.geos - DEBUG - Found GEOS DLL: <CDLL '/raid-manaslu/maref/Software/anaconda3/envs/myisceenv/lib/python3.8/site-packages/shapely/.libs/libgeos_c-a68605fd.so.1.13.1', handle 55cc50d93ba0 at 0x7f67ec742cd0>, using it.
2021-03-05 23:33:35,235 - shapely.geos - DEBUG - Trying `CDLL(libc.so.6)`
2021-03-05 23:33:35,236 - shapely.geos - DEBUG - Library path: 'libc.so.6'
2021-03-05 23:33:35,236 - shapely.geos - DEBUG - DLL: <CDLL 'libc.so.6', handle 7f6909e879d0 at 0x7f67ec753640>
Reference time:  2019-01-31 09:44:50
Satellite name:  S1B
https://qc.sentinel1.eo.esa.int/aux_poeorb/?validity_start=2019-01-29..2019-02-02&sentinel1__mission=S1B
Querying for precise orbits
Exception - something went wrong with the web scraper:
[...]

PR #254, opened 2021-03-09
Fix url change for downloading orbits

mirzaees:
Fix related to the issue #249 and #250 

Downloading Precise Orbit Ephemerides from the new url (aux.sentinel1.eo.esa.int)


rtburns-jpl:
Ah I see, the orbit downloads folder structure has also changed a bit.

Does fetchOrbit.py need to be changed too?

mirzaees:
@rtburns-jpl yes, I committed that too, could you please test to confirm that all works now?

rtburns-jpl (approved):
Nice! Works for me

PR #256, opened 2021-03-10
Convert PyCuAmpcor to pybind11

rtburns-jpl:
PyCuAmpcor is being ported into isce3, where all modules use pybind11 bindings. To make mirroring changes back and forth easier, we can port these bindings back to isce2.
We've found that for our typical style of python bindings, pybind11 is simpler and easier to use than Cython.

This needs some testing against the current PyCuAmpcor frontend scripts to ensure there are no missing bindings or overlooked bugs.


rtburns-jpl:
cc @lijun99, let me know if this looks okay to you

heavyelectron:
It looks good to me. BTW, what's the status of ISCE3? Is there a convention to follow, such as the CUDA array definitions, workflows ...?  

I am working on a revision to allow window size > 1024 (limited by the number of threads in a block in current implementation). Will update soon.

lijun99:
It looks good to me. What is the current status of ISCE3? Is there a coding convention to follow, such as the definitions of cuda arrays, workflows ... ? 

I am working on a revision to allow (oversampled) window size > 1024 ( the current implementation has this limit due to the max number of threads in a block). Will update soon.

rtburns-jpl:
Style-wise, you can look at our [.clang-format file](https://github.com/isce-framework/isce3/blob/develop/.clang-format). Our existing workflows are in [python/packages/nisar/workflows](https://github.com/isce-framework/isce3/tree/0eb175f86f6c74d86c19ba9623f23054513fc657/python/packages/nisar/workflows).

rtburns-jpl:
If no objections, planning to merge this soon before it grows stale.

lijun99:
Good to me. Thanks!
I will update the guide later to indicate the requirement for pybind11.

PR #260, opened 2021-03-17
Update dloadOrbits.py

bjmarfito:
I've added 20 additional days to reflect the date of the precise orbit file needed with respect to the image file.


bjmarfito:
I've closed this PR since https://aux.sentinel1.eo.esa.int is not available since 28 March 2021 (https://scihub.copernicus.eu/news/News00848)

PR #261, opened 2021-03-19
ALOS globals

ghost:
With cmake + GCC 10.2.1 I couldn't build due to linker errors in ALOS libraries related to duplicate symbols. I made the variables declared in headers `extern` and added a corresponding .c file. I made the symbols declared in C files `static`. Not sure if this is actually the correct fix since I didn't dig into the logic.


bhawkins-jpl:
Weird it looks like this has been fixed already, hence merge conflicts. `git pull` thinks I'm up-to-date, but it's 9 months old. Sorry for the noise.

Issue #223, opened 2021-01-27
making stackSentinel.py faster


RussellGrew:
Hi,

I have been using stackSentinel from https://github.com/isce-framework/isce2/blob/main/contrib/stack/topsStack/README.md

I noticed that most of the time was in the Snaphu phase unwrapping and that it wasn't especially CPU intensive. 

I had good outcomes manually editing the shell script to chunk runs into batches. Here batches of four.

```
# lots more chunks prior to this
SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201031_20201218 &
SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201031_20201230 &
SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201112_20201124 &
SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201112_20201206 &
wait

SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201112_20201218 &
SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201112_20201230 &
SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201112_20210111 &
SentinelWrapper.py -c /mnt/i/stack/configs/config_igram_unw_20201124_20201206 &
wait
```

I have modified `Stack.py` to accomplish this from the ` --num-process 4` command line argument. If this is of interest please let me know and I can do a pull request.  Feedback on other steps which are time consuming and could benefit is invited. I get that not all steps can be split apart into independent chunks.

Changes are in here

https://github.com/isce-framework/isce2/compare/main...RussellGrew:crude-num-process

Thanks for making this great resource available open source.

rtburns-jpl:
Thanks! Looks like a nice improvement. I'd say go ahead and submit a pull request; I'll ask if anyone else has comments/suggestions.

krasny2k5:
Very nice feature! I was thinking in doing exactly this after checking that some steps was not very CPU intensive. I'll check this and report if I find any problem.

piyushrpt:
SNAPHU runs on one core but is RAM hungry. How many unwrapping jobs one is able to run in parallel, really depends on how big the machine is. Might be better to make this an option in setting up the run files rather than hardcoding it.

RussellGrew:
Thanks for the promising feedback.

[...]

PR #271, opened 2021-04-14
stackSentinel background task fix

hardreddata:
There is a problem with https://github.com/isce-framework/isce2/issues/223 in which any tasks running in the background will continue in the background when the next step of the process begins (eg: run_03 can bleed into run_04). This causes issues when some steps try and use incomplete data. 

The proposed solution writes a final `wait` in any shell scripts which leverage `num_process`.

I welcome any feedback.


contrib/stack/topsStack/Stack.py
```diff
@@ -270,6 +270,9 @@ def write_wrapper_config2run_file(self, configName, line_cnt, numProcess = 1):
             if line_cnt == numProcess:
                 self.runf.write('wait\n\n')
                 line_cnt = 0
+            # last line of file, wait for any background processes to complete
+            if last_line:
+                self.runf.write('wait\n')
```
yunjunz, line 275: Could we merge the two `if` checks into one line?

yunjunz (commented):
Unrelated to this specific PR, but I would suggest to simplify `write_wrapper_config2run_file()` by:
+ set default `line_cnt=1`
+ change `if line_cnt == numProcess` to `if line_cnt % numProcess == 0`

In this way, the input line_cnt is the actual line index, so we don't need to return this variable, thus, shorten all calls of the func from:
```
line_cnt = configObj.write_wrapper_config2run_file(configName, line_cnt)
```
to
```
configObj.write_wrapper_config2run_file(configName, line_cnt)
```

contrib/stack/topsStack/Stack.py
```diff
@@ -358,7 +361,7 @@ def unpackSecondarysSLC(self,  stackReferenceDate, secondaryList, safe_dict):
             configObj.finalize()
 
             line_cnt += 1
-            line_cnt = configObj.write_wrapper_config2run_file(configName, line_cnt, self.numProcess)
+            line_cnt = configObj.write_wrapper_config2run_file(configName, line_cnt, secondary == secondaryList[-1], self.numProcess)
```
yunjunz, line 364: Just tiny details: I would recommend using `enumerate` such as below to eliminate the `line_cnt=0` and `line_cnt+=1` lines. 
```
for i, secondary in enumerate(secondaryList):
    configObj.write_wrapper_config2run_file(configName, line_cnt=i, numProcess=self.numProcess)
```

RussellGrew:
@yunjunz I would be happy to change

```
        line_cnt = 1
        configObj.write_wrapper_config2run_file(configName, line_cnt)
```
to
```
        configObj.write_wrapper_config2run_file(configName)
```
in the three occurrences. I left as above as it might be simpler to follow.

Presently the stackSentinel.py `num_process` is buggy so I think good to get this merged. Apologies for the delay.

contrib/stack/topsStack/Stack.py
RussellGrew: I got ahead of myself and missed some of your comments.

If we change the wrapper to default line_cnt = 1  

```
def write_wrapper_config2run_file(self, configName, line_cnt = 1, last_line = False, numProcess = 1):
```
the below works. I had to `line_cnt = i + 1` and `last_line = secondary == secondaryList[-1]` as `line_cnt` is now a keyword argument.

```
        # line_cnt = 0
        for i, secondary in enumerate(secondaryList):
            configName = os.path.join(self.config_path,'config_baseline_'+secondary)
            configObj = config(configName)
            configObj.configure(self)
            configObj.reference = os.path.join(self.work_dir,'reference/')
            configObj.secondary = os.path.join(self.work_dir,'secondarys/'+secondary)
            configObj.baselineFile = os.path.join(self.work_dir,'baselines/' + stackReferenceDate +'_' + secondary + '/' + stackReferenceDate +'_'+ secondary  + '.txt')
            configObj.computeAverageBaseline('[Function-1]')
            configObj.finalize()

            # line_cnt += 1
            configObj.write_wrapper_config2run_file(configName, line_cnt = i + 1, last_line = secondary == secondaryList[-1], numProcess = self.numProcess)
            del configObj
```

I am happy to remove the comment lines and do this in other similar locations if you like.

contrib/stack/topsStack/Stack.py
```diff
@@ -260,20 +260,19 @@ def denseOffset(self, function):
...
         # dispassionate list of commands for single process
         if numProcess == 1:
             self.runf.write(self.text_cmd + 'SentinelWrapper.py -c ' + configName + '\n')
         # aggregate background commands between wait blocks for speed gains
         elif numProcess > 1:
             self.runf.write(self.text_cmd + 'SentinelWrapper.py -c ' + configName + ' &\n')
-            if line_cnt == numProcess:
+            if line_cnt % numProcess == 0:
                 self.runf.write('wait\n\n')
                 line_cnt = 0
```
yunjunz: This line is not needed anymore

contrib/stack/topsStack/Stack.py
yunjunz: That sounds good @RussellGrew. I would suggest using the below for clarity. 

```python
last_line = secondary == secondaryList[-1]
configObj.write_wrapper_config2run_file(configName, line_cnt=i+1, last_line=last_line, numProcess=self.numProcess)

```

yunjunz:
Thank you for getting back to this @RussellGrew. Could you do the following?

1. Rebase this PR against the latest `main` branch, as the stack processor has been updated and GitHub flagged conflicts.
2. Apply the same `enumerate` style call to the new functions added within the `class run(object)`.

> @yunjunz I would be happy to change
> 
> ```
>         line_cnt = 1
>         configObj.write_wrapper_config2run_file(configName, line_cnt)
> ```
> 
> to
> 
> ```
>         configObj.write_wrapper_config2run_file(configName)
> ```
> 
> in the three occurrences. I left as above as it might be simpler to follow.

I would prefer the latter one. We don't need to show the details if we don't have to.

contrib/stack/topsStack/Stack.py
RussellGrew: I think this must have been an earlier commit. Latest version looks like `if line_cnt % numProcess == 0 or last_line:`

yunjunz: yes, you are right

RussellGrew:
@yunjunz I think that is everything.

yunjunz:
These two still stand I believe:

> 1. Rebase this PR against the latest `main` branch, as the stack processor has been updated and GitHub flagged conflicts.
> 2. Apply the same `enumerate` style call to the new functions added within the `class run(object)`. E.g., [`filtIon()`](https://github.com/isce-framework/isce2/blob/5d92004f887e16e2e78a584c6393e21b81bf157f/contrib/stack/topsStack/Stack.py#L1507). They will be visible to you after rebasing.

RussellGrew:
@yunjunz see https://github.com/isce-framework/isce2/pull/271/commits/4ffbffff0b269d13e53775394cb65da4278a1b6f and https://github.com/isce-framework/isce2/pull/271/commits/8971f5292f5803b036ac3b14e5a3893d842a2bac

Though this was the first time I have attempted to update a branch for an in progress PR. Hopefully it went ok.

RussellGrew:
That should be much closer. 

You did this to yourself with your notes of encouragement on https://mintpy.readthedocs.io/en/latest/#5_contributing

I see the CI has now passed. 

I am not familiar with these new ionosphere steps. I checked that the config is written out, and that step 17 and 18 which consider the `numProcess` are written correctly.

Cheers.

yunjunz (approved):
LGTM. Thank you @RussellGrew!

PR #272, opened 2021-04-15
cuda Scons fix: remove unnecessary -lcuda which prevents compiling on…

lijun99:
… cluster head node without cuda drivers

### DIFF
--- a/components/zerodop/GPUgeo2rdr/src/SConscript
+++ b/components/zerodop/GPUgeo2rdr/src/SConscript
@@ -3,7 +3,7 @@
 import os
 
 Import('envGPUgeo2rdr')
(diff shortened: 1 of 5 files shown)


rtburns-jpl:
Could you explain this a little more? Is it that the code doesn't need to link to the driver directly, and only uses it via cudart, so libcuda is not needed at all? I'm surprised that the node has cudart/cufft/etc but not libcuda, but if this fixes your issue then I think it looks fine to merge.

rtburns-jpl:
Okay, thanks! That makes perfect sense. Changes LGTM so I'll plan to merge this as-is.

rtburns-jpl (approved):

PR #276, opened 2021-04-23
cuDenseOffsets.prep*Geom*: supp. full-reso file w/o .full in the name

yunjunz:
This PR fixes the following potential bugs in cuDenseOffsets.py and stackStripMap.py:

+ cuDenseOffsets.prepare_geometry: support full resolution geometry files without ".full" in the filename, such as provided by stripmapStack

+ cuDenseOffsets: add --fix-xml option to update the image file path in the xml file, so that IML.loadImage is not called all the time, which could lead to file access issue when multiple processes use the same SLC file

+ stackStripMap: bugfix when --sensor is not specified. The 1st commit is for indentation adjustment following PEP8 only, please look at the 2nd commit for the actual code change.

### DIFF
--- a/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
+++ b/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
@@ -40,6 +40,8 @@ def createParser():
                         help='Reference image')
     parser.add_argument('-s', '--secondary',type=str, dest='secondary', required=True,
                         help='Secondary image')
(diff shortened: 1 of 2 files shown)


yunjunz:
Could someone please take a look at this PR?

rtburns-jpl (approved):
Sorry for the wait, LGTM. Looks like the sensor bugfix is the only functional change to the stripmap workflow. I don't know how many people are using the pycuampcor example scripts but the changes here look reasonable.

yunjunz:
Thank you for the merge @rtburns-jpl.

PR #292, opened 2021-05-17
PyCuAmpcor: add support to large windows 

lijun99:
The original code has a limit of 1024 to the (original or oversampled) secondary window size. This PR adds an implementation (cuCorrNormalizationSAT) based on sum area table(SAT) to remove that limit. Also, it adds a new cuCorrNormalizer class which selects different normalization implementations based on the window size.


rtburns-jpl (commented):
Hi, thanks for your patience with this update. I've been busy with unrelated tasks but now I have time to give this a thorough review.

Functionally, the changes look good to me. I have a few design suggestions which may simplify the code a bit:

- The cu[Corr]Normalize64/128/etc functions are identical apart from the hardcoded sizes, so we can avoid some code duplication by templating on size. Example: https://github.com/rtburns-jpl/isce2/commit/14d708863de69107bbad04b6c29c6d8f0433e8cd

- The cuNormalizer class appears to be no more than an owner for the polymorphic cuNormalizeProcessor implementation, so we can replace it with a std::unique_ptr to reuse standard ownership logic. The cuNormalizer constructor becomes a factory function which spits out a bare cuNormalizeProcessor pointer. Example: https://github.com/rtburns-jpl/isce2/commit/a0ce9d55cb39999c91bd42ca9eae305e498ddce0

N.B. I've checked that those commits compile but haven't tried running the resulting code

yunjunz (approved):
On the result side, I have been using the version of this PR for quite some pairs and it has been giving good and reasonable estimates. Thank you @lijun99!

I did not compare the result from the code before and after this PR side by side.

vbrancat (approved):
Thanks @lijun99 . I tested it on ALOS/PALSAR data and Sentinel 1A/B over fast ice and it looks good to mw.

lijun99:
@rtburns-jpl , I have merged your modifications over. I have also tested to make sure it produces the same results. Thanks a lot!

rtburns-jpl (approved):
Thanks, looks good!

PR #293, opened 2021-05-26
logging.conf: avoid empty file creation

### LINKED ISSUE #281: Zero sized isce.log gets created upon conda environment activation
Whenever my conda environment activates, it makes an empty `isce.log` file.

```bash
(mapping) scott:isce2$ mkdir emptydir && cd emptydir
(mapping) scott:emptydir$ ls
(mapping) scott:emptydir$ conda activate mapping
(mapping) scott:emptydir$ ls
isce.log
```
This means that every new window I open where my `~/.bash_profile` activates conda, I get a new file.

I haven't dug in to figure out if's from this line in the FileWriter init https://github.com/isce-framework/isce2/blob/main/components/iscesys/StdOEL/src/FileWriter.cpp#L20 or if it's just a setting for python's `logging`...  It looks like this problem was once fixed here:
https://github.com/isce-framework/isce2/pull/140
but was overwritten (possibly accidentally?) in this commit
https://github.com/isce-framework/isce2/commit/88def96714688f724913f4b8ebc926b0890055af

scottstanie:
Keeps the larger size from previous commit, but sets `delay=True` in `RotatingFileHandler`
https://docs.python.org/3/library/logging.handlers.html#logging.handlers.RotatingFileHandler

Arguments are:
`RotatingFileHandler(filename, mode='a', maxBytes=0, backupCount=0, encoding=None, delay=False, errors=None)`

### DIFF
--- a/defaults/logging/logging.conf
+++ b/defaults/logging/logging.conf
@@ -20,8 +20,8 @@ propagate=0
 [handler_fileHandler]
 class=handlers.RotatingFileHandler
 formatter=simpleFormatter
-# Filename, file mode, maximum file size in bytes,number of backups to keep
-args=('isce.log','a',1000048576,5)
+# Filename, file mode, maximum file size in bytes,number of backups to keep, encoding, delay creation
+args=('isce.log','a',1000048576,5,None,True)
 
 [handler_consoleHandler]
 class=StreamHandler


rtburns-jpl (approved):
Thanks!

PR #309, opened 2021-07-02
StAMPS preprocessor and stackprocessors minor modification

bjmarfito:
- Removed excess backslash in make_single_reference_stack_isce which inhibited the generation of files needed for StAMPS
- Removed unused imp library in topsStack and stripmapStack
- Fixed spacing on the datetime library in topsStack and stripmapStack

### DIFF
--- a/contrib/stack/stripmapStack/Stack.py
+++ b/contrib/stack/stripmapStack/Stack.py
@@ -2,7 +2,7 @@
 
 #Author: Heresh Fattahi
 
-import os, imp, sys, glob
+import os, sys, glob
 import argparse
 import configparser
 import datetime
--- a/contrib/stack/stripmapStack/invertMisreg.py
+++ b/contrib/stack/stripmapStack/invertMisreg.py
@@ -2,10 +2,10 @@
 
 # Author: Heresh Fattahi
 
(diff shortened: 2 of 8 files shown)


rtburns-jpl (approved):
LGTM, thanks!

hfattahi (approved):
LGTM

bjmarfito:
Thank you @rtburns-jpl and @hfattahi!

PeymanPython:
@hfattahi 
Many thanks for your all kind help.
Could you please introduce me the best way/tutorial to link the ISCE2StaMPS? I'v already produced the ifgs through ISCE.
Thanks in advance

bjmarfito:
@PeymanPython as far as I know, you need to use a stack of SLCs not a stack of interferograms to generate files needed for StAMPS. You must use the argument `-W slc` to generate a stack of SLCs. 

A manual was already created for ISCE2StaMPS. Here is the link: https://github.com/isce-framework/isce2-docs/blob/master/StackToStaMPS_Manual/Manual.pdf. You must take note that the terms master and slave were already discarded since ISCEv2.4 and instead it now uses the terms reference and secondary.

PR #302, opened 2021-06-14
Add lookside check to geozero, fixes #294

scottstanie:
comparison is based on https://github.com/isce-framework/isce3/blob/0eb175f86f6c74d86c19ba9623f23054513fc657/cxx/isce3/geometry/detail/Geo2Rdr.icc#L83-L89

This gets rid of the duplicate image for UAVSAR geocoding with diagonal heading

PR #318, opened 2021-07-23
bugfix in geozero and warning fix in imageMath

yunjunz:
+ geozero: bugfix for missing lookSide arg in the other call of `geozero.geozero_Py()` (#302)

+ imageMath: fix numpy deprecation warn


hfattahi (approved):
Thanks @yunjunz. This returns Geocode back to function.

PR #328, opened 2021-08-19
add "unpack" and "prepare" UAVSAR_HDF5 scripts for stripmap stack

scottstanie:
Adds the two scripts needed for [Step 4](https://github.com/isce-framework/isce2/tree/main/contrib/stack/stripmapStack#4-prepare-sar-data ) of `stackStripMap.py` to process the NISAR-simulated HDF5 products from UAVSAR. Since the `UAVSAR_HDF5_SLC` sensor was already created, I just added the prepare/extract parts so that you can do

`prepareUAVSAR_HDF5Stack.py --polarization VV --frequency A --input-dir . --output ../SLC`

I'm not sure if there's much interest in heavily adding on to ISCE2 for NISAR-like data, since that's what ISCE3 is being built for; I'm just submitting this because I've done some processing here. If it's not wanted in ISCE2, feel free to close.

### DIFF
--- a/contrib/stack/stripmapStack/prepareUAVSAR_HDF5Stack.py
+++ b/contrib/stack/stripmapStack/prepareUAVSAR_HDF5Stack.py
@@ -103,8 +103,11 @@ def main():
         slcFile = os.path.join(imgDir, imgDate + ".slc")
 
         # Now extract the correct pol SLC from the HDF5 file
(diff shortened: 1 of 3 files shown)


rtburns-jpl (approved):
Changes LGTM 👍 

I expect NISAR HDF5 products in ISCE2 will be only supported on a community basis, as mission processing needs will be met by ISCE3. But I'm happy to review and merge community contributions, and I'm sure others will find this useful.

PR #367, opened 2021-10-25
UAVSAR cropping fixes for stackStripmapApp

sssangha:
Hello team, I am working to process UAVSAR data with `stackStripmapApp.py` and ran into a few issues when attempting to crop my scenes. Below I detail the simple changes I have made to address them:

1. `frame.image.filename` and `frame.image.width` were empty in ‘cropFrames.py’, which led to crashes. Thus ‘prepareUAVSAR_coregStack.py’ was updated to populate these fields in the shelve file.

2. In `cropFrames.py`, `numpy.int` is depreciated so I replaced all `np.int` calls with `int`

3. Finally in `cropFrames.py`, I adjusted the min/max check for each dimension so only <0 differences are flagged… Important because with long inclined swaths like UAVSAR, the user may want to only crop in one dimension, and the check fails if the other isn’t cropped.

Please let me know if any further clarification/information is needed.

### DIFF
--- a/contrib/stack/stripmapStack/cropFrame.py
+++ b/contrib/stack/stripmapStack/cropFrame.py
@@ -161,18 +161,18 @@ def cropFrame(frame, limits, outdir, israw=False):
     ####sensing start
     ymin = np.floor( (limits[0] - frame.sensingStart).total_seconds() * frame.PRF)
     print('Line start: ', ymin)
-    ymin = np.int( np.clip(ymin, 0, frame.numberOfLines-1))
+    ymin = int( np.clip(ymin, 0, frame.numberOfLines-1))
 
 
     ####sensing stop 
     ymax = np.ceil( (limits[1] - frame.sensingStart).total_seconds() * frame.PRF) + 1
     print('Line stop: ', ymax)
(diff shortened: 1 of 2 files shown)


contrib/stack/stripmapStack/cropFrame.py
```diff
@@ -161,18 +161,18 @@ def cropFrame(frame, limits, outdir, israw=False):
...
     ymax = np.ceil( (limits[1] - frame.sensingStart).total_seconds() * frame.PRF) + 1
     print('Line stop: ', ymax)
-    ymax = np.int( np.clip(ymax, 1, frame.numberOfLines)) 
+    ymax = int( np.clip(ymax, 1, frame.numberOfLines)) 
 
     print('Line limits: ', ymin, ymax)
     print('Original Line Limits: ', 0, frame.numberOfLines)
 
-    if (ymax-ymin) <= 1:
+    if (ymax-ymin) < 0:
```
hfattahi, line 175: why do you want to allow zero? Does zero mean that there is one pixel overlap?

rtburns-jpl: ping @ssangha, I think this is the only part we'd like clarification about. LGTM otherwise.

PR #383, opened 2021-11-09
circleci: Match arbitrary python3 version

rtburns-jpl:
Trying to prevent the CI from barfing when the python version changes


piyushrpt:
Can also possibly automate this on more platforms by asking python3 itself. `sys.base_prefix` and `numpy.get_include()`

PR #399, opened 2021-11-24
cuDenseOffsets: bugfix in prepGeometry()

yunjunz:
This PR fixes a bug in the output geometry while running `cuDenseOffsets.py`, and applies a few more minor changes for a more user-friendly experience.

+ fix a bug of location shift in the input of `prepGeometry()`
+ add the return of `geomDict` in `estimateOffsetField()` to support re-run for geometry only without `--redo` in cmd
+ move the redo check of `cuDenseOffsets.py` back to the previous location, which is right before the actual running, in order to use the previous setup parameter code for geometry preparation only without re-run offset when it already exists.
+ remove size checking to be consistent with the latest version from Lijun with support of large win size


rtburns-jpl (approved):
LGTM

PR #410, opened 2021-12-02
Exposed an argument to update an existing stack instead of creating a new one

bjmarfito:
- Edited stackSentinel.py to expose the updateStack to enable updating of existing stack instead of creating a new one or editing line no. 428.

### DIFF
--- a/contrib/stack/topsStack/stackSentinel.py
+++ b/contrib/stack/topsStack/stackSentinel.py
@@ -445,11 +445,11 @@ def selectNeighborPairs(dateList, num_connections, updateStack):
 ########################################
 # Below are few workflow examples.
 
-def slcStack(inps, acquisitionDates, stackReferenceDate, secondaryDates, safe_dict, updateStack, mergeSLC=False):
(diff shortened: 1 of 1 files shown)


contrib/stack/topsStack/stackSentinel.py
```diff
@@ -183,6 +183,9 @@ def createParser():
 
     parser.add_argument('-rmFilter', '--rmFilter', dest='rmFilter', action='store_true', default=False,
                         help='Make an extra unwrap file in which filtering effect is removed')
+    
+    parser.add_argument('-U', '--updateStack', dest='updateStack', action='store_true', default=False,
```
hfattahi, line 187: Hi @bjmarfito . I think it might be ok to expose this parameter and force updating. But I don't see where your exposed parameter gets used?

bjmarfito: Thank you @hfattahi for the review. You are correct, I forgot to include the line of code so it can be used during processing. I will add it ASAP.

PR #420, opened 2021-12-21
runIon: import cpu-, gpu-dependent resamp funcs

yuankailiu:
The original code uses
`from isceobj.TopsProc.runFineResamp import resampSecondary`

But the runFineResamp.py has split the `resampSecondary()` function into two: `resampSecondary_CPU()` and `resampSecondary_GPU()`.

So this PR fixes this bug by identifying whether to use CPU or GPU.


rtburns-jpl (approved):
LGTM

yuankailiu:
Sorry, will this be merged into the main branch? I think it passes the review.

hfattahi (approved):
LGTM. Thank you @yuankailiu.

PR #424, opened 2021-12-30
Fix ABC alias for Python 3.10

rtburns-jpl:
Aliases in collections.* to collections.abc.* have been removed starting
with python 3.10, so we must use the fully qualified import here.

I think this is needed since https://github.com/python/cpython/pull/23754

### DIFF
--- a/components/iscesys/Component/TraitSeq.py
+++ b/components/iscesys/Component/TraitSeq.py
@@ -30,7 +30,7 @@
 
 
 
-from collections import MutableSequence
+from collections.abc import MutableSequence
 from iscesys.Component.Component import Component
 import numpy as N
 import re


vjpbd:
I got the following error when first running ISCE after installing it with Python 3.10
```
  File "/home/vincent/Logiciels/isce/v2.6.0/isce/components/iscesys/Component/TraitSeq.py", line 33, in <module>
    from collections import MutableSequence
ImportError: cannot import name 'MutableSequence' from 'collections' (/usr/lib/python3.10/collections/__init__.py)
```
I just change the aforementioned line to 
`     from collections.abc import MutableSequence`
and now it works fine.

Is it possible to add this fix to code?

PR #431, opened 2022-01-17
UAVSAR_Stack: bugfix for line/sample num and dopplerFile path

yunjunz:
This PR fixes the following two bugs in UAVSAR_Stack.py:

+ `UAVSAR_Stack`: use segment index while setting the number of lines / samples
+ `UAVSAR_Stack`: use self.dopplerFile if it exists, otherwise use the path from *.ann file, to support unpackFrame_UAVSAR.py when dop file is not in the current directory
+ `unpackFrame_UAVSAR`: rename variables and comments to improve code readability

### DIFF
--- a/components/isceobj/Sensor/UAVSAR_Stack.py
+++ b/components/isceobj/Sensor/UAVSAR_Stack.py
@@ -146,14 +146,11 @@ class UAVSAR_Stack(Component):
         frame.setSensingStart(tStart)
         frame.setSensingStop(tStop)
         frame.setSensingMid(tMid)
-        frame.setNumberOfLines(
-            int(self.metadata['slc_1_1x1_mag.set_rows']))
-        frame.setNumberOfSamples(
-            int(self.metadata['slc_1_1x1_mag.set_cols']))
+        frame.setNumberOfLines(int(self.metadata['slc_{}_1x1_mag.set_rows'.format(self.segment_index)]))
+        frame.setNumberOfSamples(int(self.metadata['slc_{}_1x1_mag.set_cols'.format(self.segment_index)]))
         frame.setPolarization(self.metadata['Polarization'])
-        frame.C0 = self.metadata['slc_1_1x1_mag.col_addr']
-        frame.S0 = self.metadata[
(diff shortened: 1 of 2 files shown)


components/isceobj/Sensor/UAVSAR_Stack.py
```diff
@@ -316,9 +312,9 @@ def extractDoppler(self):
         drho = instrument.getRangePixelSize()  #full res value, not spacing in the dop file
         prf = instrument.getPulseRepetitionFrequency()
         self.logger.info("extractDoppler: rho0, drho, prf = {}, {}, {}".format(rho0, drho, prf))
-        dopfile = self.metadata['dop']
-        f = open(dopfile,'r')
-        x = f.readlines()  #first line is a header
+        dopfile = self.dopplerFile if hasattr(self, 'dopplerFile') else self.metadata['dop']
```
rtburns-jpl, line 315: instead of hasattr, you can use getattr with the default value fallback

contrib/stack/stripmapStack/unpackFrame_UAVSAR.py
```diff
@@ -1,61 +1,64 @@
...
-        os.mkdir(slcname)
+    if not os.path.isdir(slcDir):
+        os.mkdir(slcDir)
 
-    pickName = os.path.join(slcname, 'data')
+    pickName = os.path.join(slcDir, 'data')
     with shelve.open(pickName) as db:
         db['frame'] = obj.frame
 
+    return
```
rtburns-jpl, line 50: bare return is not necessary

rtburns-jpl (commented):
Looks great, just a couple small nits

components/isceobj/Sensor/UAVSAR_Stack.py
yunjunz: Done. Thanks for the tip!

contrib/stack/stripmapStack/unpackFrame_UAVSAR.py
yunjunz: This is removed now.

rtburns-jpl (approved):
Looks great!

PR #441, opened 2022-02-15
Fix denseoffsetsRead for modern gfortran

rtburns-jpl:
Allows compilation without need for -fallow-argument-mismatch

My intent here is to declare getLineBand's `arr` argument as `type(*)`, which as far as I can tell is the fortran equivalent of a `void*` and allows passing any data. I'm a fortran newbie so I would appreciate a look from someone more experienced.

### DIFF
--- a/.cmake/isce2_buildflags.cmake
+++ b/.cmake/isce2_buildflags.cmake
@@ -9,11 +9,6 @@ add_compile_options(
     $<$<COMPILE_LANGUAGE:Fortran>:-ffree-line-length-none>
     $<$<COMPILE_LANGUAGE:Fortran>:-fno-range-check>
     $<$<COMPILE_LANGUAGE:Fortran>:-fno-second-underscore>)
-if(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU" AND
-   CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
-    add_compile_options(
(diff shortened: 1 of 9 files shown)


EJFielding:
In this case, I don't think that long Fortran experience is any help. I learned Fortran 77 and then some of the Fortran 90 and Fortran 95 additions. I found this: `Fortran 2018 added TYPE(*) or assumed type as part of the enhanced C interoperability features` (https://stevelionel.com/drfortran/2020/06/30/doctor-fortran-in-not-my-type/) so it is very recent. It sounds like it is what we need here to accept a variable passed back from C code, but I have never seen it before.

rtburns-jpl:
Thanks for pointing that out, and I completely agree that the lack of backward compatibility is annoying. I'm going to think about this some more and try to come up with something that is more broadly compatible.

rtburns-jpl:
I was looking at some other geospatial packages and I think I found a solution that doesn't require recent gfortran versions.

https://github.com/ecmwf/eccodes/commit/2e0d31080dda0bdd06ae2e1d1d2037498c85e020

Instead of trying to teach fortran that a function is polymorphic, we can just make a bunch of aliases for any different type of argument we might want to pass, and have them all call the same underlying C function.

EJFielding (approved):
This seems like a more reliable and easier to understand way of handing the different types than expecting several versions of the Fortran compiler to figure it out correctly. Thanks for taking the time to implement a better version.

rtburns-jpl:
Thanks for your valuable feedback :)

PR #442, opened 2022-02-16
Switch CI base images to ubuntu 20.04

rtburns-jpl:


rtburns-jpl:
The lack of an opencv-devel package on Oracle Linux 8 might be a blocker here. Going to try switching to Ubuntu 20.04, which is also NASA-approved. cc @pymonger

PR #505, opened 2022-06-17
absolute module import in `stackSentinel/StripMap.py`

yunjunz:
This PR replaces the `from Stack import *` line in `stackSentinel.py` and `stackStripMap.py` with the absolute module import below, to allow for pythonic calls to their `main()`. 

```
from topsStack.Stack import *
from stripmapStack.Stack import *
```

This change requires adding `$ISCE_STACK` to `$PYTHONPATH`, thus, would break the previous default setup for stack processors. But it enables a more programmable workflow/recipe for the downstream developers. So I think it's a good change at a small cost: updating the stack processor path setup, which is updated in the README file.

The PR also includes a small change on `alosStack`: use $ISCE_STACK/alosStack to replace $PATH_ALOSSTACK for style consistency

### DIFF
--- a/contrib/stack/README.md
+++ b/contrib/stack/README.md
@@ -1,37 +1,76 @@
-## Stack Processors
+## ISCE-2 Stack Processors
 
 Read the document for each stack processor for details.
 
-+ [stripmapStack](./stripmapStack/README.md)
 + [topsStack](./topsStack/README.md)
++ [stripmapStack](./stripmapStack/README.md)
 + [alosStack](./alosStack/alosStack_tutorial.txt)
 
 ### Installation
 
 To use a stack processor you need to:
 
-1. Install ISCE as usual
+#### 1. Install ISCE as usual
 
-2. Depending on which stack processor you need to try, add the path of the folder containing the python scripts to your `$PATH` environment variable as follows:
-   - add the full path of your **contrib/stack/topsStack** to `$PATH` to use the topsStack for processing a stack of Sentinel-1 TOPS data
-   - add the full path of your **contrib/stack/stripmapStack** to `$PATH` to use the stripmapStack for processing a stack of StripMap data
-   - set environment variable `$PATH_ALOSSTACK` by doing: export PATH_ALOSSTACK=CODE_DIR/contrib/stack/alosStack to use the alosStack for processing a stack of ALOS-2 data
+#### 2. Setup paths for stack processors
 
-Note: The stack processors do not show up in the install directory of your isce software. They can be found in the isce source directory. 
+The stack processors do not show up in the install directory of your isce software. They can be found in the isce source directory. Thus, extra path setup is needed.
+
+2.1 Add the following path to your `${PYTHON_PATH}` environment vavriable:
+
+```bash
+export ISCE_STACK={full_path_to_your_contrib/stack}
+export PYTHONPATH=${PYTHONPATH}:${ISCE_STACK}
+```
+
+2.2 Depending on which stack processor you want to use, add the following path to your `${PATH}` environment variable:
+
++ For Sentinel-1 TOPS data
+
+```bash
+export PATH=${PATH}:${ISCE_STACK}/topsStack
+```
+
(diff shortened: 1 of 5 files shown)


hfattahi (approved):
LGTM!

EJFielding:
Looks helpful to make the `alosStack` setup more similar to the `topsStack` and `stripmapStack` setup.

rtburns-jpl:
> This change requires adding `$ISCE_STACK` to `$PYTHONPATH`

What if we also install topsStack, stripmapStack, etc along with the other python modules? Then we can simply `import isce.components.contrib.topsStack` instead of having to modify environment variables.

EJFielding:
@rtburns-jpl I think the problem is that `topsStack` and `stripmapStack` have command-line programs with the same name but slightly different functions. That is why @hfattahi kept them separate and we have to add one or the other to the `$PYTHONPATH` and `$PATH` to do the processing workflow.

yunjunz:
If installed via conda, all the stack processors are currently included in the `envs/*/share/isce2` folder, whose path is saved as `$ISCE_STACK`. I don't know why they are not installed to `/envs/insar/lib/python3.8/site-packages/isce` directory, whose path is saved as `$ISCE_HOME`, which is accessible via `import isce.*`. 

Since topsStack is more frequently used in its development version (for me at least), is there an easy way to point `isce.components.contrib.topsStack` to a non-interpreter directory?

yunjunz:
Although the current way of adding `$ISCE_STACK` to `$PYTHON_PATH` from this PR is not ideal (as described in @rtburns-jpl comments above), it's consistent with the current setup style of adding `$ISCE_STACK` to the `$PATH`. I would suggest that we merge this PR before the new release.

EJFielding:
We should decide today or tomorrow whether to merge this before we freeze version for the UNAVCO 2022 short course. I don't understand enough about how Python works to advise on the details.

rtburns-jpl:
> This change requires adding $ISCE_STACK to $PYTHONPATH, thus, would break the previous default setup for stack processors.

If I understand that correctly, we definitely shouldn't merge something breaking for a bugfix release. I think we can come up with something that will not break existing setups, and will make the stack more programmable without involving more env variable trickery.

yunjunz:
> If I understand that correctly, we definitely shouldn't merge something breaking for a bugfix release. 

Yeah, I agree with this.

yunjunz:
> I think we can come up with something that will not break existing setups, and will make the stack more programmable without involving more env variable trickery.

I would argue that:
1. the existing setups are manual already, this PR would bring one more line of manual setup, plus a few simplifications to improve the style consistency, thus, it's clear and simple to adjust on the user side, in terms of breaking change; 
2. in the logic of "functionality is more important than elegant implementation", the programmable workflow (which is vital in my own processing) is still worth merging before we find out a better way, which would require other changes that may deserve its own PR.

Thus, I would think it is good to be merged, probably after the currently planned next bugfix release. 

Adding @hfattahi and @piyushrpt to see if they have opinions on this.

hfattahi:
@rtburns-jpl I think this is useful functionality and would be nice to merge. You can increment the minor release and make it clear in README of the release what users need to change to be able to keep running the stack processor. @yunjunz can help with that clarification.

rtburns-jpl:
Ok sounds good, thanks. I'll include this in the next feature release.

yunjunz:
Thanks, @hfattahi and @rtburns-jpl. Here is a note for clarification in the release:

NOTE: Since this release, users should add `${ISCE_STACK}` to their `${PYTHON_PATH}` to be able to keep running the stack processors. For `bash` users, e.g., it's: `export PYTHONPATH=${PYTHONPATH}:${ISCE_STACK}`. Check more detailed instruction in the [README](https://github.com/isce-framework/isce2/blob/main/contrib/stack/README.md) file.

EJFielding:
@rtburns-jpl When you have some time, I think this is a substantial change to ISCE2 that would merit at least a minor version bump.

PR #516, opened 2022-07-04
Tutorial fix to avoid confusion for alos2App

shitong01:
Fix typo to ALOS2 tutorial to avoid confusion about ionospheric corrected inteferogram's filenames. Please do correct me if my understanding of the files are wrong.

Thanks!


CunrenLiang:
Thanks, Shi Tong. We should fix this typo @rtburns-jpl.

rtburns-jpl:
Thanks for the fix!

PR #558, opened 2022-08-10
Changed SRTM download to `https`

EJFielding:
LP DAAC now requires `https` and refuses the `http` connections.

PR #560, opened 2022-08-10
Changed SRTM download to `https`

rtburns-jpl:
cc @EJFielding (https://github.com/isce-framework/isce2/pull/558)

### DIFF
--- a/contrib/demUtils/demstitcher/DemStitcherND.py
+++ b/contrib/demUtils/demstitcher/DemStitcherND.py
@@ -44,7 +44,7 @@ from iscesys.Component.Component import Component
 from contrib.demUtils.DemStitcher import DemStitcher as DS
 #Parameters definitions
 URL1 = Component.Parameter('_url1',
-    public_name = 'URL1',default = 'http://e4ftl01.cr.usgs.gov/MEASURES/NASADEM_HGT.001/2000.02.11',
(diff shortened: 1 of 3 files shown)


EJFielding:
Thanks for the clean PR! Can you go ahead and merge this? I tested a `dem.py` download and it now works.

EJFielding:
For some reason, I can download the URLs with `wget` from a command line, but the `curl` command that is run from inside `dem.py` is connecting but not getting any data. They should be using the same `$HOME/.netrc` file.

EJFielding:
The `dem.py` run did finally finish after about 15 minutes, so the URL is correct but the servers still have some capacity issues.

kylemurray2:
I had to add the 'https' to iscesys/DataManager/SWBDManager.py get the watermask working.  There might be a few other URLs that need to be changed.

EJFielding:
@kylemurray2 Thanks for spotting that other place that the URLs are hard-coded. I did a search of the whole directory tree and found some more in `DataManager` and other locations. See #562

Xin-Wang-520:
> For some reason, I can download the URLs with `wget` from a command line, but the `curl` command that is run from inside `dem.py` is connecting but not getting any data. They should be using the same `$HOME/.netrc` file.

Teacher I also have this problem, after changing the address, but still can not download, how to solve？

PR #561, opened 2022-08-10
Update release_history.py

EJFielding:
Updated release history for Feb. 2022 2.6.0 release.


rtburns-jpl:
Looks good, but please drop the whitespace changes. (They are an improvement but should be in a separate commit or PR)

EJFielding:
Sorry, I opened the file in `Atom` and it did that automatically.

PR #573, opened 2022-08-26
bugfix to properly install envisat.so

bjmarfito:
Changes in this code to fix the error showed below.

`OSError: /home/bryanjim001/.conda/envs/isce2expe/lib/python3.9/site-packages/isce/components/isceobj/Sensor/envisat.so: cannot open shared object file: No such file or directory`

The code asa_im_decode.c should be installed as envisat.so so it can be recognized by ENVISAT.py.


rtburns-jpl (approved):
Good catch, thanks

Issue #581, opened 2022-09-13
Can't find s5 directory when running alos2burstApp.py


bjmarfito:
I encountered the error below when I ran alos2burstApp.py.

```
Polynomial Order: 2 - by - 2
0.0185615	-3.18133e-06	2.97508e-09
-1.15301e-05	-2.92791e-09	0
2.8038e-09	0	0
Traceback (most recent call last):
  File "/home/bryanjim001/.conda/envs/isce2expe/bin/alos2burstApp.py", line 1098, in <module>
    insar.run()
  File "/home/bryanjim001/.conda/envs/isce2expe/lib/python3.9/site-packages/isce/components/iscesys/Component/Application.py", line 142, in run
    exitStatus = self._processSteps()
  File "/home/bryanjim001/.conda/envs/isce2expe/lib/python3.9/site-packages/isce/components/iscesys/Component/Application.py", line 405, in _processSteps
    result = func(*pargs, **kwargs)
  File "/home/bryanjim001/.conda/envs/isce2expe/lib/python3.9/site-packages/isce/components/isceobj/Alos2burstProc/Factories.py", line 40, in __call__
    return self.method(self.other, *args, **kwargs)
  File "/home/bryanjim001/.conda/envs/isce2expe/lib/python3.9/site-packages/isce/components/isceobj/Alos2burstProc/runCoregCc.py", line 42, in runCoregCc
    os.chdir(swathDir)
FileNotFoundError: [Errno 2] No such file or directory: 's5'
```

bjmarfito:
This error occurs in step coreg_cc.

EJFielding:
Did you include all five subswaths in the processing parameters?

bjmarfito:
Yes @EJFielding, I included them on the xml file.

```
<property name="starting swath">1</property>
<property name="ending swath">5</property>
```

bjmarfito:
Does the warning below affect the coreg_cc step?

```
WARNING: land area too small for estimating offsets between reference and secondary magnitudes at frame 3250, swath 4
set offsets to zero
```

bjmarfito:
Hi @EJFielding, I have fixed this issue by adding a os.chdir('../') after line no. 71 on the runCoregCc.py.

I will submit a PR for this and if you have time you can check if it works properly.

CunrenLiang:
ALOS-2 still needs cross correlaiton to estimate offsets between reference and secondary images. Better not include subswaths without land or very little land.

[...]

PR #582, opened 2022-09-13
Added os.chdir('../') after line no. 71

bjmarfito:
Added os.chdir('../') after line no. 71 so it can change directory to the next swath after printing the warning message (pls. see this issue https://github.com/isce-framework/isce2/issues/581).


bjmarfito:
This PR doesn't solve the issue. I managed to get pass on coreg_cc but the next step produces an error.

PR #596, opened 2022-10-01
Added lines in downsampleDEM.py to tag downsampled DEMs as WGS84 and corrected misspelled height in alos2App application, modules, and examples

bjmarfito:
- Added code lines adapted from https://github.com/scottstanie/sardem/blob/master/sardem/utils.py in downsampleDEM.py to tag downsampled DEMs as WGS84 ellipsoid


bjmarfito:
Corrected misspelled height in alos2App application, modules, and examples

PR #597, opened 2022-10-04
Merge similar cuArraysCopy helpers using templates

rtburns-jpl:
This is to ease C++-only conversion of the cuArraysCopy kernels by reducing code duplication

cc @lijun99 if you would like to apply this patch to https://github.com/lijun99/cuampcor

### DIFF
--- a/contrib/PyCuAmpcor/src/cuAmpcorUtil.h
+++ b/contrib/PyCuAmpcor/src/cuAmpcorUtil.h
@@ -29,21 +29,16 @@ void cuArraysCopyC2R(cuArrays<float2> *image1, cuArrays<float> *image2, int stri
 
 // same routine name overloaded for different data type
 // extract data from a large image
-void cuArraysCopyExtract(cuArrays<float> *imagesIn, cuArrays<float> *imagesOut, cuArrays<int2> *offset, cudaStream_t stream);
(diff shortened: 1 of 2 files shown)


gmgunter (approved):
LGTM :+1:

I suppose a possible small downside of this new approach is that you will get a link-time error rather than a compile-time error if you try to invoke any of these templated functions with an unsupported type. But this seems unlikely to be an issue in practice, and anyway I think the benefits of reducing copy-paste by consolidating these functions far outweigh this minor potential disadvantage.

lijun99:
Thanks, @rtburns-jpl ! When the code was started, the template was not supported by cuda. Many things have changed. I was thinking to rewrite the array with mdspan or Michael's grid. 

The adoption of template looks good to me. But  haven't tried the code yet. One quick question, when calling template kernels,  is adding the template specifications no longer needed, e.g.,  <T_in, T_out> after cuArraysCopyPadded_kernel<T_in, T_out> ?

rtburns-jpl:
> One quick question, when calling template kernels, is adding the template specifications no longer needed, e.g., <T_in, T_out> after cuArraysCopyPadded_kernel<T_in, T_out> ?

Yes, that's right. When the template types can be auto-determined from the parameters passed to the function, it isn't necessary to explicitly specify the template types.

PR #639, opened 2022-12-19
Remove depricated np.complex alias in ImageLib

jhkennedy:
`np.complex` has been deprecated since v1.20 and has now been removed in NumPy v1.24:
https://numpy.org/doc/stable/release/1.24.0-notes.html

> The deprecation for the aliases `np.object`, `np.bool`, `np.float`, `np.complex`, `np.str`, and `np.int` is expired (introduces NumPy 1.20). Some of these will now give a FutureWarning in addition to raising an error since they will be mapped to the NumPy scalars in the future.
> 
> ([gh-22607](https://github.com/numpy/numpy/pull/22607))

This prevents [TopsApp.py](https://github.com/isce-framework/isce2/blob/bb4d3b545b0400a9c15e5dd7b6f1cb577f77bfa2/applications/topsApp.py#L46) from being used:

```shell
$ conda create -n tmp isce2 numpy=1.24
$ conda activate tmp
$ python -c ' import isce; from isceobj import TopsProc'
This is the Open Source version of ISCE.
Some of the workflows depend on a separate licensed package.
To obtain the licensed package, please make a request for ISCE
through the website: https://download.jpl.nasa.gov/ops/request/index.cfm.
Alternatively, if you are a member, or can become a member of WinSAR
you may be able to obtain access to a version of the licensed sofware at
https://winsar.unavco.org/software/isce
  File "<string>", line 1, in <module>
  File "$HOME/mambaforge/envs/tmp/lib/python3.11/site-packages/isce/components/isceobj/TopsProc/__init__.py", line 7, in <module>
    from .Factories import *
  File "$HOME/mambaforge/envs/tmp/lib/python3.11/site-packages/isce/components/isceobj/TopsProc/Factories.py", line 81, in <module>
    createSubsetOverlaps = _factory("runSubsetOverlaps")
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "$HOME/mambaforge/envs/tmp/lib/python3.11/site-packages/isce/components/isceobj/TopsProc/Factories.py", line 15, in _factory
    module = __import__(
             ^^^^^^^^^^^
  File "$HOME/mambaforge/envs/tmp/lib/python3.11/site-packages/isce/components/isceobj/TopsProc/runSubsetOverlaps.py", line 14, in <module>
    from isceobj.Util.ImageUtil import ImageLib as IML
  File "$HOME/mambaforge/envs/tmp/lib/python3.11/site-packages/isce/components/isceobj/Util/ImageUtil/ImageLib.py", line 70, in <module>
    "J"   : np.complex(0.0, 1.0),
            ^^^^^^^^^^
  File "$HOME/mambaforge/envs/tmp/lib/python3.11/site-packages/numpy/__init__.py", line 284, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'complex'. Did you mean: 'complex_'?
```

### DIFF
--- a/components/isceobj/Util/ImageUtil/ImageLib.py
+++ b/components/isceobj/Util/ImageUtil/ImageLib.py
@@ -67,8 +67,8 @@ fnDict = { 'cos':       np.cos,
 
 #######Current list of constants
 constDict = { "PI"  : np.pi,
-              "J"   : np.complex(0.0, 1.0),
-              "I"   : np.complex(0.0, 1.0),
+              "J"   : np.complex64(0.0, 1.0),
+              "I"   : np.complex64(0.0, 1.0),
               "E"   : np.exp(1.0),
               "NAN" : np.nan,
               "ROW" : None,
--- a/components/isceobj/Util/ImageUtil/ImageLib.py
+++ b/components/isceobj/Util/ImageUtil/ImageLib.py
@@ -67,8 +67,8 @@ fnDict = { 'cos':       np.cos,
 
 #######Current list of constants
 constDict = { "PI"  : np.pi,
-              "J"   : np.complex64(0.0, 1.0),
-              "I"   : np.complex64(0.0, 1.0),
+              "J"   : 1j,
+              "I"   : 1j,
               "E"   : np.exp(1.0),
               "NAN" : np.nan,
               "ROW" : None,


components/isceobj/Util/ImageUtil/ImageLib.py
```diff
@@ -67,8 +67,8 @@
 
 #######Current list of constants
 constDict = { "PI"  : np.pi,
-              "J"   : np.complex(0.0, 1.0),
-              "I"   : np.complex(0.0, 1.0),
+              "J"   : np.complex64(0.0, 1.0),
+              "I"   : np.complex64(0.0, 1.0),
```
rtburns-jpl, line 71: Having trouble finding a migration guide for this change, but based on this line in the cpython PR, should this be complex128? https://github.com/numpy/numpy/pull/22607/files#diff-09a2c2e86cca91dac0c336e4dad47197ba3cd837ff6c85abf8525cc3b879d3e6L181

rtburns-jpl, line 71:
```suggestion
              "J"   : 1j,
              "I"   : 1j,
```

Alternatively, can we just use python complex literals, or do these need to be numpy types?

jhkennedy: @rtburns-jpl I think python complex literals should work

jhkennedy: @rtburns-jpl I committed your `1j` suggestion, but that'll (I think) be interpreted by `numpy` as `complex128` instead of `complex64`. I think that's likely fine, but if we want it to stay `complex64,` we'll need to explicitly do that.

jhkennedy:
```
              "J"   : np.complex64(1j),
              "I"   : np.complex64(1j),
```
would do it

jhkennedy: So, since originally this would have been `complex128` via the old numpy alias, I'm content to leave it with the python native type

rtburns-jpl (approved):
Thanks for looking into this! LGTM

PR #641, opened 2022-12-22
Updated parser to use stripmapApp conventions for Doppler centroids

### LINKED ISSUE #640: Poor focusing using the ROI_PAC sensor type in stripmapApp
This is an edge case for sure, but I am trying to process legacy JERS data from raw using stripmapApp. Since there is no example input file format (that I could find in the current version, anyway) for JERS, I am trying to use the ROI_PAC input format. I did once successfully process this pair with ROI_PAC, so this was pretty easy to set up, with the raw files, rsc files and all that.

The results, however, were disappointing. Neither image is well focused at the SLC stage, and everything beyond that is (as you might expect) fuzzy garbage. I am left wondering if there is something in the metadata parsing that did not make it from ROI_PAC into ISCE? I note that both the raw files have significant Doppler shifts (-1.2 PRF in the ROI_PAC nomenclature) which does make me wonder if that might be the issue? Anyone have any suggestions?

fuzzy_garbage

geniusinaction:


piyushrpt (commented):
I would suggest adding roi_pac sensor type to list that uses default doppler method as well.

geniusinaction:
Added ROI_PAC to the default list, per suggestion. Fixed an indent.

piyushrpt (approved):
LGTM!!!

rtburns-jpl (approved):
Looks great, I'll go ahead and merge. I liked your twitter thread on this, btw :)

PR #661, opened 2023-02-16
Replace deprecated numpy types with native types

rtburns-jpl:
Closes https://github.com/isce-framework/isce2/issues/660

cc @yuankailiu for testing

PR #669, opened 2023-02-23
Replace deprecated `normed` with `density` in numpy.histogram()

yuankailiu:
Related to #661, expired deprecations from numpy 1.24.

The normed keyword argument has been removed from np.histogram, np.histogram2d, and np.histogramdd. Use density instead. If normed was passed by position, density is now used.

Here is the change log:
https://numpy.org/devdocs/release/2.24.1-notes.html#expired-deprecations


rtburns-jpl:
Good catch, thanks!

PR #695, opened 2023-04-10
`wbd`: add `-u/--url` option to allow for different urls

yunjunz:
This PR adds the `wbd.py -u/--url` option to be able to change the water body tile files' url, similar to the `dem.py --url` option. The default behavior is kept the same, so no change is expected on the user side or in the previous usage. Detailed changes are as below:

+ `isceobj.Alos2Proc.runDownloadDem.download_wbd()`: add `url` optional argument to be able to change the default url for the water body tile files.

+ `applications.wbd.py`:
   - add `cmd_line_parse()` function to handle the command line inputs.
   - add `-u/--url` option to allow changing the url of water body tile files.
   - add example usage

### DIFF
--- a/applications/wbd.py
+++ b/applications/wbd.py
@@ -6,9 +6,41 @@
 #
 
 
+import argparse
 import sys
 import isce
 from isceobj.Alos2Proc.runDownloadDem import download_wbd
+
+
+EXAMPLE = """Usage examples:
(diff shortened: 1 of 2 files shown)


rtburns-jpl (approved):
Looks good, example snippets run successfully. Thanks for the argparse cleanup, will be much more maintainable 👍

PR #709, opened 2023-05-04
bugfix in `cuDenseOffsets.py --full/out-geom` for `topsApp` product

yunjunz:
This PR fixes a bug while running `cuDenseOffsets.py --full/out-geom` on `topsApp.py` products by:
+ setting `lat` as the first geometry dataset, instead of `hgt`, because the former is shared between topsApp and topsStack products
+ add `z` from topsApp, in addition to `hgt` from topsStack, as the input geometry dataset

### DIFF
--- a/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
+++ b/contrib/PyCuAmpcor/examples/cuDenseOffsets.py
@@ -394,7 +394,7 @@ def estimateOffsetField(reference, secondary, inps=None):
 
 
 def prepareGeometry(full_dir, out_dir, x_start, y_start, x_step, y_step, x_win_num, y_win_num,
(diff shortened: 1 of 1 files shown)


bjmarfito:
Fixed the problem that I'm having with topsApp height products being processed through `cuDenseOffsets.py`.

bjmarfito:
Hi @rtburns-jpl, can this PR be merged since I have coregistered SLC generated by topsApp and I will use PyCuAmpCor to generate pixel offsets? Thank you!

rtburns-jpl:
Thanks for confirming the fix, will go ahead and merge it. Sorry for the wait!

bjmarfito:
Thank you for merging!

PR #721, opened 2023-05-23
Suggest to update fetchOrbit.py and topsStack README.md

yuankailiu:
### fetchOrbit.py: 

+ Allow reading SLC zip files from a given folder via `-d` option. So fetchOrbit.py can be used directly as a script to download a batch of orbit files by specifying the SLC folder, rather than doing `-i` for each input *.zip file.

### topsStack README.md:
+ The run_files descriptions in [`README.md` of the topsStack processor](https://github.com/isce-framework/isce2/blob/main/contrib/stack/topsStack/README.md) are outdated. Here the proper names of each run file is used. Some brief explanations are also updated for each run file.

### DIFF
--- a/contrib/stack/topsStack/README.md
+++ b/contrib/stack/topsStack/README.md
@@ -6,9 +6,9 @@ The detailed algorithm for stack processing of TOPS data can be find here:
 
 -----------------------------------
 
-To use the sentinel stack processor, make sure to add the path of your `contrib/stack/topsStack` folder to your `$PATH` environment varibale. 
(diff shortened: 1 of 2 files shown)


chintals:
Great edit! One brief comment (not related to this PR), the commands in run files for ionospheric estimation are all in serial order (except run_ns+1, subband_and_resamp) regardless of nproc setting. Curious as to why the ns+ files are not parallelized with nproc setting?

yuankailiu:
I don't know. I have always been using HPC cores to do the parallelization of each command line in every run file with `stackSentinel.py --num_proc 1`.

PR #622, opened 2022-11-15
PyCuAmpcor: compile files as pure C++ when possible, fix Makefile

rtburns-jpl:
cc @lijun99

PR #722, opened 2023-05-25
PyCuAmpcor: offer an option to fix the starting pixel locations

lijun99:
Some changes were made to allow users to fix the coordinate of the first window (center). This will be useful in cases such as 1) to compare the results with different window sizes (center needs to be fixed); 2) to find out the offsets for a particular pixel or a group of pixels.  

When the center is fixed, it is possible that some pixels in the window are out of range when the window is close to borders. Zeros are padded for these out-of-range pixels. 

The code tidying from #622 is also merged. Thanks, @rtburns-jpl !


lijun99:
A recent update [51af91d](https://github.com/isce-framework/isce2/pull/722/commits/51af91d211ca6be7b2cf3c645bda3f5b1eaadeb2) fixes the issue of gpu memory leak when repeating the cuAmpcor processes within the same python script.

hfattahi:
@lijun99 sorry for getting to this PR very late. We have had enough with the ISCE3, NISAR etc. Is this PR still the latest or do you want to change anything? 
@vbrancat and @rtburns-jpl please take a look at this PR. Also, do we need to port any part to isce3?

lijun99:
Sorry for my late reply, I was into other things as well.  

Please consider approving this PR, which includes two major updates 
1) an option to choose the location of the starting pixel, which is good for debugging, also, comparing the offsets with different window sizes; 
2) a memory leak issue if the script is called repeatedly.  

Thanks! @hfattahi  @rtburns-jpl  @vbrancat

lijun99:
Now included in PR#922.

PR #735, opened 2023-06-26
fix bug in topo.py in stripmapStack

yjzhenglamarmota:
When determining whether multi-looking geometry files is needed

 - should be inps.alks * inps.rlks >1 , not inps.rlks * inps.rlks >1


yunjunz (approved):

rtburns-jpl:
Thanks!

PR #753, opened 2023-08-09
Changed mlab.griddata to scipy.interpolate.griddata 

bjmarfito:
- Changed mlab.griddata to scipy.interpolate.griddata on  runDispersive.py and estimateIono.py since mlab.griddata was deprecated in matplotlib.


rtburns-jpl (approved):
I don't understand how this ever worked in the first place since there's no mlab import, so let's go ahead and merge this :)

PR #767, opened 2023-09-10
unpackFrame_ROIPAC_raw.py fixes

bjmarfito:
- Use default values instead of using DOPIQ when focusing images
- Use `os.path.abspath` for the image name so stackStripMap.py can read the path of the raw files when processing
- Change the file format to UNIX using `:set ff=unix`

### DIFF
--- a/contrib/stack/stripmapStack/unpackFrame_ROIPAC_raw.py
+++ b/contrib/stack/stripmapStack/unpackFrame_ROIPAC_raw.py
@@ -5,6 +5,8 @@ from isceobj.Sensor import createSensor
 import shelve
 import argparse
 import os
+from mroipac.dopiq.DopIQ import DopIQ
+import copy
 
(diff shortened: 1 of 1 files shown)


hfattahi:
@bjmarfito was there a reason that you remove the Doppler estimation here? 
I doubt if there are Roi_pac users. Not possible for me to test right now. Can you confirm if you have tested and with what kind of data?

bjmarfito:
Hi Heresh, I used the unpackFrame_ROIPAC_raw.py to process ROIPAC formatted JERS data. I used the default doppler estimation for JERS data and it worked correctly. I actually can't remember the reason why I removed the DOPIQ since it was many months ago after I submitted this PR. Though, I already brought it back for users who need it and added an argument to either use DEFAULT of DOPIQ for doppler estimation.

PR #834, opened 2024-03-27
`alosStack`: easy data preparation + ignore non-numeric folders in the data/download dir

yunjunz:
This PR adds two features to facilitate the `alosStack` processing. Default behaviors are not changed.

+ `stripmapStack/prepSlcALOS2.py`: add the optional `--alosStack` option to leverage this script to prepare the SLC data structure expected by `alosStack`. Default is turned off.

+ `alosStack`: ignore the non-numeric folders in the data download directory, similar to stripmapStack, such as the `ARCHIVED_FILES` folder generated by `stripmapStack/prepSlcALOS2` above, instead of (re)moving the original downloaded zip/tar files. This includes changes in the following scripts:
   - StackPublic
   - create_cmds
   - read_data

It also includes two minor print out msg and bug fixes:

+ `alosStack/create_cmds`: print out the number of acquisitions and pairs

+ `alosStack/create_cmds`: ignore empty input of `insar.bbox` arg, in addition to the existing None input, otherwise, it causes the following error:

```bash
  File "/home/yunjunz/tools/isce2/src/isce2/contrib/stack/alosStack/geocode.py", line 72, in <module>
    bbox = [float(x) for x in bbox.split('/')]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yunjunz/tools/isce2/src/isce2/contrib/stack/alosStack/geocode.py", line 72, in <listcomp>
    bbox = [float(x) for x in bbox.split('/')]
            ^^^^^^^^
ValueError: could not convert string to float: ''
```

### DIFF
--- a/contrib/stack/alosStack/StackPulic.py
+++ b/contrib/stack/alosStack/StackPulic.py
@@ -187,7 +187,7 @@ def stackDateStatistics(idir, dateReference):
 
     #get date folders
     dateDirs = sorted(glob.glob(os.path.join(os.path.abspath(idir), '*')))
-    dateDirs = [x for x in dateDirs if os.path.isdir(x)]
+    dateDirs = [x for x in dateDirs if os.path.isdir(x) and os.path.basename(x).isdigit()]
 
     #find index of reference date:
     dates = []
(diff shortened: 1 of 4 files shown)


hfattahi (approved):
LGTM. The original functionality seems untouched and the new functionality is turned off by default.

PR #877, opened 2024-09-02
Snaphu update

EJFielding:
These small changes to `IsFinite` function were necessary to compile on arm64 (Apple Silicon).


bjmarfito:
This was also fixed by a more comprehensive PR https://github.com/isce-framework/isce2/pull/897/files which addressed more issues in newer GCC versions.

EJFielding:
Ah, I will close my PR since the fixes were applied.

PR #886, opened 2024-09-27
Update README.md for topsStack ionospheric phase screen module

bjmarfito:
- Corrected three lines related to ionospheric phase screen outputs to be consistent with other list within the group


rtburns-jpl:
Thanks, the surrounding lines also had some extraneous tab characters so I removed those as well.

PR #897, opened 2024-11-07
fix some incorrect pointer types as required by newer gcc compilers

lijun99:
I fixed some code which are reported as "incorrect pointer/type" error by newer versions of gcc. The updated code can be compiled by Apple Clang as well. 

@CunrenLiang, please make sure the changes don't break the code.

### DIFF
--- a/contrib/Snaphu/include/snaphu.h
+++ b/contrib/Snaphu/include/snaphu.h
@@ -141,11 +141,11 @@
 #define DEF_NLOOKSAZ         5
 #define DEF_NLOOKSOTHER      1
 #define DEF_NCORRLOOKS       23.8
-#define DEF_NCORRLOOKSRANGE  3  
+#define DEF_NCORRLOOKSRANGE  3
 #define DEF_NCORRLOOKSAZ     15
 #define DEF_NEARRANGE        831000.0
 #define DEF_DR               8.0
-#define DEF_DA               20.0 
+#define DEF_DA               20.0
 #define DEF_RANGERES         10.0
 #define DEF_AZRES            6.0
 #define DEF_LAMBDA           0.0565647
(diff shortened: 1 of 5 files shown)


lijun99:
Sorry, my editor prefers to remove all the empty spaces in the end. The actual changes are not that many, 
1) snaphu.h, snaphu_utils.c, change the IsFinite function to int type, and use isfinite function from math.h; 
2) contrib/alos2filter/src/psfilt1.c change the 
   p_forward  = fftwf_plan_dft_2d(fftw, fftw, (fftw_complex *)seg_fft[0], (fftw_complex *)seg_fft[0], FFTW_FORWARD, FFTW_MEASURE);
   p_backward = fftwf_plan_dft_2d(fftw, fftw, (fftw_complex *)seg_fft[0], (fftw_complex *)seg_fft[0], FFTW_BACKWARD, FFTW_MEASURE);
to 
   p_forward  = fftwf_plan_dft_2d(fftw, fftw, (fftwf_complex *)seg_fft[0], (fftwf_complex *)seg_fft[0], FFTW_FORWARD, FFTW_MEASURE);
   p_backward = fftwf_plan_dft_2d(fftw, fftw, (fftwf_complex *)seg_fft[0], (fftwf_complex *)seg_fft[0], FFTW_BACKWARD, FFTW_MEASURE);
since fftw_complex is of double type. 
also changed the IsFinite function; 
3) graphx_mdx.c, enforce a type conversion for XmProcessTraversal: it returns a bool, but the XtCallbackProc should be a void type.

bjmarfito:
Tested on Linux and macOS with GCC 14 and it works. Thank you @lijun99!

Can this be tested and merged @rtburns-jpl since we have a new system and we can only use GCC 14 and newer? Thanks!

PR #946, opened 2025-04-03
Two patches for scons to run properly

lijun99:
When using scons install method with python 13 and gcc/gfortan 13.3, I ran into two issues 
1) an error with getLineBand in denseoffsetsRead.F and estimateoffsetsRead.F.  The solution is to use the respective getLineBand method for real or complex data, as defined in  components/iscesys/ImageApi/DataAccessor/src/DataAccessorF.cpp. 
(these filed are not included in the CMake file, are they no longer in use?) 
2) warnings issued during PyCuAmpcor compilation, "Two environments with different actions...". Basically, all the functions are included in the pybind11 module and we don't need an additional lib (which causes the two different actions).

### DIFF
--- a/components/isceobj/Util/denseoffsets/src/denseoffsetsRead.F
+++ b/components/isceobj/Util/denseoffsets/src/denseoffsetsRead.F
@@ -39,7 +39,7 @@
                     integer*8 :: acc
                     integer :: irow,band,n,i
 
-                    call getLineBand(acc,carr,band,irow)
+                    call getLineBand_c(acc,carr,band,irow)
                 end subroutine readCpxAmp
 
                 subroutine readAmp(acc,arr,irow,band,n,carr)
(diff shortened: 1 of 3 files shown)


jeffrey1102:
I am having problems with the scons installation, it gives me different errors and asks for license

lijun99:
For those who want to get scons working before this PR is merged, please use my repo at first, 

    git remote add lijun99  https://github.com/lijun99/isce2.git
    git fetch lijun99
    git checkout patch

jeffrey1102:
Your Repo works and scons ends the installation.
When you go back to your guide and making the isce2.mod, the command module load doesn’t work even if I install the (enviroment-module) the command is not found. I also did it with the bash but it also says that it doesn’t find isce in the path
how do you proceed after your RP?

Issue #947, opened 2025-04-04
Encounter an issue when processing Sentinel-1C


wpfeng:
Dear all,

I had an issue when processing  Sentinle-1C TOPS SAR with current isce2 using the stack strategy. The error information is posted below...

....
Setting IPF version to :  003.91
Encountered unknown mission id S1C

After information showing up, there is no further step to read orbit data and extract target bursts... :( 

Is there any advice to solve this issue asap?

Cheers,
Wanpeng

geniusinaction:
You have to add a couple of lines to one of the processing scripts ($ISCE_HOME/components/isceobj/Sensor/TOPS/Sentinel1.py)

The relevant code block is after line 602. I added another entry for S1C:

```
if mission == 'S1A':
                burst.trackNumber = (orbitnumber-73)%175 + 1
            elif mission == 'S1B':
                burst.trackNumber = (orbitnumber-27)%175 + 1
            elif mission == 'S1C':
                burst.trackNumber = (orbitnumber+3)%175 + 1
            else:
                raise ValueError('Encountered unknown mission id {0}'.format(mission))
```

The tricky bit was converting the orbit number to the track number. I think I got it, but it would be great if someone could check my working, and maybe push a fix!

wpfeng:
Dear  Funning,

Thanks for your solution. Yes.  This is cool. We have also figured out with the same way after a discussion with Cunren.

Thanks a lot anyway.

Best regards,
Wanpeng

bjmarfito:
Also tested it and worked on our side. Thanks @geniusinaction

PR #951, opened 2025-04-18
Support Sentinel-1C

AndrewPlayer3:
This PR adds the relative orbit calculation for Sentinel-1C so that `ValueError('Encountered unknown mission id S1C` is not thrown. 

Note that this is the same formula as used in [s1-reader](https://github.com/isce-framework/s1-reader/pull/144/files).

This PR should address issue #947.


asjohnston-asf:
@rtburns-jpl you appear to have handled all the merges to `main` in the last year, are you in a position to review this PR?

rtburns-jpl (approved):
Thanks for the ping, and sorry about the delay - looks good to me and happy to merge this!

PR #982, opened 2025-08-25
ALOS-4 reader for stripmapApp.py and SAOCOM-1 parser for the stripmap stack processor

fdelgadodelapuente:


fdelgadodelapuente:
Hi,

I added a parser for ALOS-4 stripmap data. It is a modification of the ALOS-2 code with minimal modifications to properly handle the new ALOS-4 data.  For the moment I tested only with stripmapApp.py, so it does not include ionospheric corrections and no support for ScanSAR.

I also uploaded a parser for  SAOCOM-1 stripmap data for the stack processor. Just like ALOS-2 it does not include corrections for the dispersive phase. I used it in this paper
https://ieeexplore.ieee.org/document/10586971

 

Francisco

